PR #9, opened 2017-11-23
Add support for distributing to multiple distribution groups

johannth:
This PR adds support for releasing a build to multiple distribution groups using a comma separated list in the `group` parameter.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -313,7 +313,10 @@ module Fastlane
           if uploaded
             release_url = uploaded['release_url']
             UI.message("Release committed")
-            self.add_to_group(api_token, release_url, group, release_notes)
+            groups = group.split(',')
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -497,7 +500,7 @@ def self.available_options
 
           FastlaneCore::ConfigItem.new(key: :group,
                                   env_name: "APPCENTER_DISTRIBUTE_GROUP",
-                               description: "Distribute group name",
+                               description: "Comma separated list of distribute group names",
```
derpixeldan: Can you please change the description to `Comma separated list of Distribution Group names`. I know it was wrong before already, but why not doing it here and now :)

evkhramkov:
@johannth thanks for your contribution, I picked your changes and added unit tests in #13

johannth:
Thanks for the feedback @derpixeldan and #13 looks great @evkhramkov. Should I then just close this PR?

PR #14, opened 2017-11-29
Better API Token link

janpio:
direct link to generation of api token instead of linking to docs that explain how to test the api via swagger (although it's nice this is possible!)


lumaxis:
Thanks @janpio, nice catch and good improvement! 🙂 👍

PR #24, opened 2017-12-11
Get full release info, when adding it to group

### LINKED ISSUE #23: lane_context is wrongly filled
Hi,

Thanks for this plugin! 👍 

I've found a bug, due to [this patch HTTP request](https://github.com/Microsoft/fastlane-plugin-appcenter/blob/master/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb#L206). In `release = response.body` is always just `"distribution_group_name", "release_notes"` which is probably unwanted because I would expect the whole release object.

I'm willing to open pull request, just don't know if my solution is good enough. https://github.com/jukben/fastlane-plugin-appcenter/commit/95f6375567454932941c409983fc08efe9c34f47

evkhramkov:
fix for #23


jukben:
Good job! I've also wanted about encapsulate the logic into a self-sustain function. Nicer and cleaner than my hotfix. 👍 

LGTM.

PR #32, opened 2018-04-10
Adds Destination parameter

leoMehlig:
This add a Destination names parameter which can be used to submit builds automatically to a store.
Example:
```
appcenter_upload(
      app_name: "AppName",
      destination: "iTunes Connect users"
    )
```
This will release every build to TestFlight.


lumaxis:
@leoMehlig Looks like there are still two minor Rubocop violations. Could you take a look and fix them too? Thanks!

leoMehlig:
@lumaxis Fixed them!

PR #34, opened 2018-05-21
Fixed crash

RishabhTayal:
I was getting following error:

`no implicit conversion of Symbol into Integer`

Calling code:

```
params = {
      api_token: 'xxxxxxxxx',
      upload_dsym_only: true,
      owner_name: 'owner',
      app_name: app_name,
      dsym: 'path to dsym',
      release_notes: 'test'
    }
Fastlane::Actions::AppcenterUploadAction.run_dsym_upload(params)
```

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -325,8 +325,6 @@ module Fastlane
             UI.message("dSYM files zipped")
           end
 
-          values[:dsym_path] = dsym_path
-
           UI.message("Starting dSYM upload...")
           dsym_upload_details = self.create_dsym_upload(api_token, owner_name, app_name)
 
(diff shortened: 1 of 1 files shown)


RishabhTayal:
I am not sure if this is the right fix for the crash I am having. However it fixes the crash for me.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -325,8 +325,6 @@ def self.run_dsym_upload(params)
             UI.message("dSYM files zipped")
           end
 
-          values[:dsym_path] = dsym_path
```
evkhramkov: This part is used only in tests, so I'm not sure why it caused a crash for you.

Instead of calling Action method `run_dsym_upload` consider using following:
```ruby
appcenter_upload({
  api_token: 'xxx',
  owner_name: 'owner',
  app_name: 'app',
  upload_dsym_only: true,
  dsym: './spec/fixtures/symbols/Themoji.dSYM.zip'
})
```

evkhramkov:
Closing, because crash caused by improper action usage

PR #38, opened 2018-10-16
Added notify_testers

c0un7-z3r0:
Adds the ability to decide if testers should be notified.
Also fixes rubocop config

Example
```
appcenter_upload({
          api_token: 'xxx',
          owner_name: 'owner',
          app_name: 'app',
          apk: './spec/fixtures/appfiles/apk_file_empty.apk',
          group: 'Testers',
          notify_testers: false
        })
```


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -219,7 +219,7 @@ def self.get_release(api_token, release_url)
       end
 
       # add release to distribution group
-      def self.add_to_group(api_token, release_url, group_name, release_notes = '')
+      def self.add_to_group(api_token, release_url, group_name, release_notes = '', notify_testers = true)
```
lumaxis: I would personally vote for the default here being `false` since that makes more sense to me in a CI scenario like where the fastlane plugin is likely used.
However, I believe this would be a breaking change to the current default behavior without specifying the flag explicitly which does always trigger the notification.

spec/appcenter_upload_spec.rb
```diff
@@ -272,6 +272,53 @@ def stub_add_to_group(status)
...
+    it "should notify users" do
+      stub_check_app(200)
+      stub_create_release_upload(200)
+      stub_upload_build(200)
+      stub_update_release_upload(200, 'committed')
+      stub_add_to_group(200)
+      stub_get_release(200)
+
+      values = Fastlane::FastFile.new.parse("lane :test do
+        Actions.lane_context[SharedValues::FL_CHANGELOG] = 'autogenerated changelog'
```
lumaxis: Is this line needed in the context of this spec?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
c0un7-z3r0: I think so too, currently it always triggers the notification. Which is why I did not want to change the behaviour. But I also see your point in the CI case. How would do you guys handle this? I'm open for both.

spec/appcenter_upload_spec.rb
c0un7-z3r0: True, I removed it.

evkhramkov:
Added as part of #57

PR #41, opened 2018-11-06
Fix crash when printing doc

AliSoftware:
Running `fastlane action appcenter_upload` to print the documentation of the plugin/action crashes fastlane, because `self.details` is supposed to be a `String`, not an `Array`.

This commit fixes `self.details` to return a string instead of an array, which will prevent `fastlane action` to crash.


AliSoftware:
Not sure why Travis fails, doesn't seem related to the changes I made?

ceyhuno:
#45 Should fix Travis

lumaxis:
Thanks to @ceyhuno, CI is now fixed and we're finally able to merge this 🙏🏼 Cheers @AliSoftware !

PR #62, opened 2019-04-14
Allow specifying app os, platform and display name for automated creation

### LINKED ISSUE #58: Add optional parameters app OS, platform & display name
According to user feedback the app slug alone isn't helpful to identify the app and obviously any prompt breaks CI.

Proposal:
* Add optional parameters `app_os`, `app_platform`, `app_display_name` which will be used if the app slug not found
* If all parameters provided, there will be no prompts, so it will be possible to use the action in CI

References:
https://github.com/Microsoft/fastlane-plugin-appcenter/issues/17 https://github.com/Microsoft/fastlane-plugin-appcenter/issues/19 https://github.com/Microsoft/fastlane-plugin-appcenter/issues/52

evkhramkov:
Added following optional parameters
`app_display_name`(if not specified, will use `app_name` value as default)
`app_os`
`app_platform`

If this parameters specified, upload action will automatically create new app if it was not found without prompting user.
This will simplify plugin usage in CI environments.

Closes #58 

Demo:
```ruby
appcenter_upload(
  api_token: ENV["TEST_APPCENTER_API_TOKEN"],
  owner_name: ENV["TEST_APPCENTER_OWNER_NAME"],
  app_name: "my-new-app-center-app",
  app_display_name: "My New App Center App",
  app_os: "Android",
  app_platform: "Java",
  apk: "./fastlane/app-release.apk"
)
```

Output:

Screen Shot 2019-04-14 at 10 00 26 AM


evkhramkov:
> it doesn't create an app if it was found

@jphenow Yes it is covered. In all test cases where it is assumed that an application is found, there is no stub for creating an application request - therefore, if it is caused, the test fails due to an unregistered request

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -121,20 +121,31 @@ def self.get_or_create_app(params)
...
+        
         if Helper::AppcenterHelper.get_app(api_token, owner_name, app_name)
           true
         else
-          if Helper.test? || UI.confirm("App with name #{app_name} not found, create one?")
-            os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
-            platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
-
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+          should_create_app = !app_display_name.to_s.empty? || !app_os.to_s.empty? || app_platform.to_s.empty?
```
DennisPan: Should this be `!app_platform.to_s.empty?` instead of `app_platform.to_s.empty?`?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -189,13 +200,31 @@ def self.available_options
 
           FastlaneCore::ConfigItem.new(key: :app_name,
                                   env_name: "APPCENTER_APP_NAME",
-                               description: "App name. If there is no app with such name, you will be prompted to create one",
+                               description: "App slug. If there is no app with such slug, you will be prompted to create one",
```
DennisPan: This is not the slug. `<owner_name>/<app_name>` is the slug

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -189,13 +200,31 @@ def self.available_options
...
                                   optional: false,
                                       type: String,
                               verify_block: proc do |value|
                                 UI.user_error!("No App name given, pass using `app_name: 'app name'`") unless value && !value.empty?
                               end),
 
+          FastlaneCore::ConfigItem.new(key: :app_display_name,
+                                  env_name: "APPCENTER_APP_DISPLAY_NAME",
+                               description: "App display name. Used for new app creation, if app with 'app_name' slug was not found",
```
DennisPan: Sane here. `app_name` is not slug

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: Oh, thank you for note

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: yes, thank you for notice

fastlane/Fastfile
```diff
@@ -16,6 +16,19 @@ lane :test_autocreation do
   )
 end
 
+lane :test_autocreation_without_prompts do
```
mseyma: just curious, if it did ask for prompts how does the test fail?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -134,20 +134,31 @@ def self.get_or_create_app(params)
...
+        
         if Helper::AppcenterHelper.get_app(api_token, owner_name, app_name)
           true
         else
-          if Helper.test? || UI.confirm("App with name #{app_name} not found, create one?")
-            os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
-            platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
-
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+          should_create_app = !app_display_name.to_s.empty? || !app_os.to_s.empty? || !app_platform.to_s.empty?
```
mseyma: is there ever a case where the app is created but one of those fields is empty?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -134,20 +134,31 @@ def self.get_or_create_app(params)
...
 
         platforms = {
           "Android" => ['Java', 'React-Native', 'Xamarin'],
           "iOS" => ['Objective-C-Swift', 'React-Native', 'Xamarin']
         }
 
+        
         if Helper::AppcenterHelper.get_app(api_token, owner_name, app_name)
           true
         else
```
mseyma: nit since the if conditional above returns, we could reduce nesting and remove the else clause if we wanted

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -134,20 +134,31 @@ def self.get_or_create_app(params)
...
-          if Helper.test? || UI.confirm("App with name #{app_name} not found, create one?")
-            os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
-            platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
-
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+          should_create_app = !app_display_name.to_s.empty? || !app_os.to_s.empty? || !app_platform.to_s.empty?
+          
+          if Helper.test? || should_create_app || UI.confirm("App with name #{app_name} not found, create one?")
```
mseyma: nit we could reduce nesting if we wanted by rewriting like:
```
if !Helper.test? && !should_create_app && !UI.confirm("App with name #{app_name} not found, create one?")
    UI.error("Lane aborted")
    false

app_display_name = app_name if app_display_name.to_s.empty?
...
...
```

spec/appcenter_upload_spec.rb
```diff
@@ -6,8 +6,11 @@ def stub_check_app(status)
     )
 end
 
-def stub_create_app(status)
+def stub_create_app(status, app_name = "app", app_display_name = "app", app_os = "Android", app_platform = "Java")
```
mseyma: do we want the default `app_display_name` value to be different from `app_name` so that we can verify that `app_name` isn't used in place of `app_display_name`? same goes for test below

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -134,20 +134,31 @@ def self.get_or_create_app(params)
...
+        app_display_name = params[:app_display_name]
+        app_os = params[:app_os]
+        app_platform = params[:app_platform]
 
         platforms = {
           "Android" => ['Java', 'React-Native', 'Xamarin'],
           "iOS" => ['Objective-C-Swift', 'React-Native', 'Xamarin']
         }
 
+
```
ranterle: 🙃 Extra white space necessary here?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -134,20 +134,31 @@ def self.get_or_create_app(params)
...
         else
-          if Helper.test? || UI.confirm("App with name #{app_name} not found, create one?")
-            os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
-            platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
-
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+          should_create_app = !app_display_name.to_s.empty? || !app_os.to_s.empty? || !app_platform.to_s.empty?
+          
+          if Helper.test? || should_create_app || UI.confirm("App with name #{app_name} not found, create one?")
+            app_display_name = app_name if app_display_name.to_s.empty?
```
ranterle: 🙃 maybe `app_display_name = app_name unless app_display_name.to_s.empty` is more readable

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -209,6 +220,24 @@ def self.available_options
                                 UI.user_error!("No App name given, pass using `app_name: 'app name'`") unless value && !value.empty?
                               end),
 
+          FastlaneCore::ConfigItem.new(key: :app_display_name,
+                                  env_name: "APPCENTER_APP_DISPLAY_NAME",
+                               description: "App display name. Used for new app creation, if app with 'app_name' name was not found",
```
ranterle: 🙃 Potentially "App display name to use when creating a new app" is simpler.
The logic that it creates a new app if app with `app_name` is not found might also be added to that config items description.

fastlane/Fastfile
evkhramkov: In code we have additional checks `Helper.test?`, which returns `true` in test env. So all prompts skipped, e.g.:
```rb
if Helper.test? || should_create_app || UI.confirm("App with name #{app_name} not found, create one?")
```

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: but it is opposite condition, isn't it? we need to set display name to name if parameter is empty

spec/appcenter_upload_spec.rb
evkhramkov: Personally I'm not sure that we need that

cc @derpixeldan any thoughts?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: 👍 I agree, thanks!

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
ranterle: I think you're right - got mixed in double negatives I guess.

spec/appcenter_upload_spec.rb
DennisPan: I think @mseyma was asking in the context of the test. I think that's a valid suggestion

derpixeldan: I think it's a valid suggestion for the test scenario.

evkhramkov: Oh, I got it. I already have a separate test especially for that, so it is covered

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: this `else` with `UI.error("Lane aborted")` message is for case when user didn't accept `UI.confirm("App with name #{app_name} not found, create one?")`

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: if not all parameters configured, it will go to prompt

Issue #64, opened 2019-04-17
invalid distribution group name


emilioicai:
I recently updated to 0.2.1 . Everything seemed to be working fine until I tried distributing my app to the Testflight and Beta testers in Google play. Now I only get this error over and over:

```
Not found, invalid distribution group name
Group 'Alpha' was not found
```

This is th relevant code in my Fastfile:

```
# upload to appcenter
appcenter_upload(
      api_token: "<TOKEN>",
      owner_name: "<OWNER>",
      app_name: "<APPNAME>",
      notify_testers: true,
      group: "Alpha",
      release_notes: "<NOTES>",
      apk: lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
)
```

Which makes me think something changed in the way appcenter retrieves the group id. Can you confirm this is the case? If so, how should I modify my Fastfile to reflect this change?

emilioicai:
I can confirm the distribution group 'Alpha' exists in my App's appcenter setup

evkhramkov:
Hello, sorry for that issue. At the moment, plugin supports distributing only to App Center distribution groups, you can’t target external destination (like TestFlight, or Google Play)

emilioicai:
This was working before, any plans on supporting this feature in the future?

evkhramkov:
Yes we will fix that soon, I hope we'll be able to release an update next week.
But due to API changes, which were adopted in `0.2.0` release, we'll have to introduce separate parameter to specify external stores.

Sorry again for that issue, but you still can try downgrading to `0.1.7`

DennisPan:
My thoughts on this is to first refactor what we have and change `group` to `destinations`, and add `destination_type`, which is consistent with how App Center CLI does this (https://github.com/Microsoft/appcenter-cli/blob/master/src/commands/distribute/releases/add-destination.ts#L67) and App Center public APIs. Does this sound reasonable @evkhramkov ?

evkhramkov:
@emilioicai we just released version `1.0.0` with stores support. You can use that like that:
```ruby
# upload to appcenter
appcenter_upload(
      api_token: "<TOKEN>",
      owner_name: "<OWNER>",
      app_name: "<APPNAME>",
      notify_testers: true,
      destinations: "Alpha",
      destination_type: "store",
      release_notes: "<NOTES>",
      apk: lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
)
```

[...]

PR #73, opened 2019-05-02
Refactor to support destination type, in preparation of supporting distributing to store

DennisPan:
First step of address #64 

1. Support `destinations` and `destination_type` as the inputs. Both are optional and default to `Collaborators` distribution group.
2. Only `group` is supported as `destination_type`. Support for others will be added in future PRs.
3. Deprecate `group` input. Throws error when it is used.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -275,7 +275,7 @@ module Fastlane
                                   optional: true,
                                       type: String,
                               verify_block: proc do |value|
-                                UI.user_error!("No or incorrect destination type given. Use `destination_type: 'group'`") unless value && !value.empty? && value == "group"
+                                UI.user_error!("No or wrong destination type given. Use `destination_type: 'group'`") unless value && !value.empty? && value == "group"
                               end),
 
           FastlaneCore::ConfigItem.new(key: :mandatory_update,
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -253,10 +253,31 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :group,
                                   env_name: "APPCENTER_DISTRIBUTE_GROUP",
                                description: "Comma separated list of Distribution Group names",
+                                  optional: true,
+                                      type: String,
+                                deprecated: true,
+                              verify_block: proc do |value|
+                                UI.user_error!("Option `group` is depreated. Use `destinations` and `destination_type`")
```
mseyma: nit typo `depreated` (deprecated?)

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -253,10 +253,31 @@ def self.available_options
...
 
+
+          FastlaneCore::ConfigItem.new(key: :destination_type,
+                                  env_name: "APPCENTER_DISTRIBUTE_DESTINATION_TYPE",
+                               description: "Destination type of distribution destination. Currently only 'group' is supported",
+                             default_value: "group",
+                                  optional: true,
+                                      type: String,
+                              verify_block: proc do |value|
+                                UI.user_error!("No or wrong destination type given. Use `destination_type: 'group'`") unless value && !value.empty? && value == "group"
```
mseyma: nit maybe incorrect instead of wrong?

PR #75, opened 2019-05-16
support autocreating apps in org

evkhramkov:
Transfer created app to specified owner if needed (when owner is org)

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -149,7 +149,7 @@ module Fastlane
 
             created = Helper::AppcenterHelper.create_app(api_token, app_name, os, platform)
             # by default app created for authorized user
-            # we need to update owner if it action called for org
+            # we need to update owner if action called for org
             if created && created['owner']['name'] != owner_name
               Helper::AppcenterHelper.transfer_app(api_token, created['owner']['name'], app_name, owner_name)
             end
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -147,7 +147,12 @@ module Fastlane
             os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
             platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
 
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+            created = Helper::AppcenterHelper.create_app(api_token, app_name, os, platform)
+            # by default app created for authorized user
+            # we need to update owner if it action called for org
+            if created && created['owner']['name'] != owner_name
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -147,7 +147,12 @@ def self.get_or_create_app(params)
             os = Helper.test? ? "Android" : UI.select("Select OS", ["Android", "iOS"])
             platform = Helper.test? ? "Java" : UI.select("Select Platform", platforms[os])
 
-            Helper::AppcenterHelper.create_app(api_token, owner_name, app_name, os, platform)
+            created = Helper::AppcenterHelper.create_app(api_token, app_name, os, platform)
+            # by default app created for authorized user
+            # we need to update owner if it action called for org
```
lumaxis: Typo?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -147,7 +147,12 @@ def self.get_or_create_app(params)
...
+            if created && created['owner']['name'] != owner_name
+              Helper::AppcenterHelper.transfer_app(api_token, created['owner']['name'], app_name, owner_name)
```
lumaxis: Sorry if this is a naive question but why don't we just create the app under the correct owner right away? 🤔
Because we don't know if the specified `owner_name` is the current user or a different org?

evkhramkov: Yes, you're right. We can add a separate parameter `toOrg`, but I'm not sure that this will be useful.
Auto app creation is something that called just once per app, so I don't see big issue in additional request, instead of adding additional parameter.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: Is that about `... if it action ...`?

lumaxis: Yes! 😄

evkhramkov: thanks, fixed!

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
DennisPan: This seems pretty odd to me too. We can add a `owner_type` input, which defaults to `user`, but can be changed to `organization`. I would suggest to make this something that is more explicitly called out by customers, not that we implicitly do some magic for them. We can use some PM input here too @derpixeldan

derpixeldan: The question is, how error prone is the "magic" flow. The only edge case I can think of at the moment is that a user has the same name as an org he owns. In that case they won't be able to specify whether the app is created in the user or in the org and we have to default to either?

zanizrules:
Hey guys, I was wondering if a decision had been made? I require this functionality to tick some boxes on a project and would prefer not to branch off from master.

evkhramkov:
Implemented proper logic in #104, closing this one

PR #76, opened 2019-05-20
Support distributing to store

### LINKED ISSUE #64: invalid distribution group name

DennisPan:
Closes #64


lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -245,26 +245,26 @@ def self.get_release(api_token, owner_name, app_name, release_id)
...
           req.headers['internal-request-source'] = "fastlane"
         end
 
         case response.status
         when 200...300
-          group = response.body
-          UI.message("DEBUG: received group #{JSON.pretty_generate(group)}") if ENV['DEBUG']
-          group
+          destination = response.body
+          UI.message("DEBUG: received #{destination_type} #{JSON.pretty_generate(group)}") if ENV['DEBUG']
```
evkhramkov: Debug message should be fixed from `#{JSON.pretty_generate(group)}"` to `#{JSON.pretty_generate(destination)}"`

evkhramkov: That will fix tests

DennisPan: Nice catch!

PR #84, opened 2019-06-06
Add android mapping support

bryansmi:
Adds official mapping support.
Warns that unofficial mapping support is being deprecated.
Renames internal functions for symbol_upload to be more generic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,10 @@ The action parameters `api_token` and `owner_name` can also be omitted when thei
 - `APPCENTER_APP_NAME` - App name. If there is no app with such name, you will be prompted to create one
 - `APPCENTER_DISTRIBUTE_APK` - Build release path for android build
 - `APPCENTER_DISTRIBUTE_IPA` - Build release path for ios build
-- `APPCENTER_DISTRIBUTE_DSYM` - Path to your symbols file. For iOS provide path to app.dSYM.zip
+- `APPCENTER_DISTRIBUTE_DSYM` - Path to your symbols (app.dSYM.zip) file 
 - `APPCENTER_DISTRIBUTE_UPLOAD_DSYM_ONLY` - Flag to upload only the dSYM file to App Center
+- `APPCENTER_DISTRIBUTE_MAPPING` - Path to your Android mapping.txt file 
+- `APPCENTER_DISTRIBUTE_UPLOAD_MAPPING_ONLY` - Flag to upload only the mapping file to App Center
 - `APPCENTER_DISTRIBUTE_DESTINATIONS` - Comma separated list of destination names. Both distribution groups and stores are supported. All names are required to be of the same destination type. Default is `Collaborators`.
 - `APPCENTER_DISTRIBUTE_DESTINATION_TYPE` - Destination type of distribution destination. `group` and `store` are supported. Default is `group`
 - `APPCENTER_DISTRIBUTE_MANDATORY_UPDATE` - Require users to update to this release
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ The action parameters `api_token` and `owner_name` can also be omitted when thei
 - `APPCENTER_DISTRIBUTE_IPA` - Build release path for ios build
 - `APPCENTER_DISTRIBUTE_DSYM` - Path to your symbols (app.dSYM.zip) file 
 - `APPCENTER_DISTRIBUTE_UPLOAD_DSYM_ONLY` - Flag to upload only the dSYM file to App Center
-- `APPCENTER_DISTRIBUTE_MAPPING` - Path to your Android mapping.txt file 
-- `APPCENTER_DISTRIBUTE_UPLOAD_MAPPING_ONLY` - Flag to upload only the mapping file to App Center
+- `APPCENTER_DISTRIBUTE_ANDROID_MAPPING` - Path to your Android mapping.txt file 
+- `APPCENTER_DISTRIBUTE_UPLOAD_ANDROID_MAPPING_ONLY` - Flag to upload only the mapping file to App Center
 - `APPCENTER_DISTRIBUTE_DESTINATIONS` - Comma separated list of destination names. Both distribution groups and stores are supported. All names are required to be of the same destination type. Default is `Collaborators`.
 - `APPCENTER_DISTRIBUTE_DESTINATION_TYPE` - Destination type of distribution destination. `group` and `store` are supported. Default is `group`
 - `APPCENTER_DISTRIBUTE_MANDATORY_UPDATE` - Require users to update to this release
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -278,7 +278,7 @@ module Fastlane
                                 deprecated_files = ["mapping.txt"]
                                 if value
                                   UI.user_error!("Couldn't find dSYM file at path '#{value}'") unless File.exist?(value)
-                                  UI.message("Support for mapping.txt has been deprecated. Please use APPCENTER_DISTRIBUTE_ANDROID_MAPPING instead.") if deprecated_files.include? File.basename(value)
+                                  UI.message("Support for mapping.txt has been deprecated. Please use --mapping parameter or APPCENTER_DISTRIBUTE_ANDROID_MAPPING environment variable instead.") if deprecated_files.include? File.basename(value)
                                 end
                               end),
 
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -78,7 +78,7 @@ module Fastlane
         end
 
         UI.message("Starting mapping upload...")
-        symbol_upload_details = Helper::AppcenterHelper.create_mapping_upload(api_token, owner_name, app_name, file, build_number, version)
+        symbol_upload_details = Helper::AppcenterHelper.create_mapping_upload(api_token, owner_name, app_name, mapping, build_number, version)
 
         if symbol_upload_details
           symbol_upload_id = symbol_upload_details['symbol_upload_id']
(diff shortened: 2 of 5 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -258,6 +289,26 @@ def self.available_options
                                  is_string: false,
                              default_value: false),
 
+          FastlaneCore::ConfigItem.new(key: :mapping,
+                                  env_name: "APPCENTER_DISTRIBUTE_ANDROID_MAPPING",
+                               description: "Path to your Android Proguard or R8 mapping.txt",
+                                  optional: true,
+                                      type: String,
+                              verify_block: proc do |value|
+                                accepted_formats = ["mapping.txt"]
```
evkhramkov: Let's not restrict to only `mapping.txt` input

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -246,8 +275,10 @@ def self.available_options
                                   optional: true,
                                       type: String,
                               verify_block: proc do |value|
+                                deprecated_files = ["mapping.txt"]
```
evkhramkov: I think it make sense to check any `.txt` file

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
ranterle: Agree, when we talked about this last, this was mostly in the context of a CI build running, or uploading the results of a CI build, in which case the file we're looking for is named "mapping.txt" - but in the case of both CLI and Fastlane (see comment in https://github.com/microsoft/appcenter-cli/pull/594 as well), it is possible to run the upload from a different location. 
So we should not restrict the file name for the mapping file to only mapping.txt - we can still check for the .txt extension. Also, when uploading to Diagnostics, we can specify the filename - it would be great if we can use the base name of the mapping file here, so this does not confuse users.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -57,11 +58,37 @@ def self.run_dsym_upload(params)
...
+
+        UI.message("Starting mapping upload...")
+        symbol_upload_details = Helper::AppcenterHelper.create_mapping_upload(api_token, owner_name, app_name, file, build_number, version)
+
+        if symbol_upload_details
+          symbol_upload_id = symbol_upload_details['symbol_upload_id']
+          upload_url = symbol_upload_details['upload_url']
+
+          UI.message("Uploading mapping...")
+          Helper::AppcenterHelper.upload_symbol(api_token, owner_name, app_name, mapping, "Android", symbol_upload_id, upload_url)
```
ranterle: I don't see where the symbol name "Android" is mapped to the value "AndroidProguard" that App Center Diagnostics API requires for Android mappings, maybe I'm missing it.

README.md
```diff
@@ -36,8 +36,10 @@ The action parameters `api_token` and `owner_name` can also be omitted when thei
 - `APPCENTER_APP_NAME` - App name. If there is no app with such name, you will be prompted to create one
 - `APPCENTER_DISTRIBUTE_APK` - Build release path for android build
 - `APPCENTER_DISTRIBUTE_IPA` - Build release path for ios build
-- `APPCENTER_DISTRIBUTE_DSYM` - Path to your symbols file. For iOS provide path to app.dSYM.zip
+- `APPCENTER_DISTRIBUTE_DSYM` - Path to your symbols (app.dSYM.zip) file 
 - `APPCENTER_DISTRIBUTE_UPLOAD_DSYM_ONLY` - Flag to upload only the dSYM file to App Center
+- `APPCENTER_DISTRIBUTE_MAPPING` - Path to your Android mapping.txt file
```
ranterle: Maybe this should say "Android mapping file" here instead

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -258,6 +289,26 @@ def self.available_options
                                  is_string: false,
                              default_value: false),
 
+          FastlaneCore::ConfigItem.new(key: :mapping,
+                                  env_name: "APPCENTER_DISTRIBUTE_ANDROID_MAPPING",
+                               description: "Path to your Android Proguard or R8 mapping.txt",
```
ranterle: I think we should not reference ProGuard or R8 here – if we do, please spell it "ProGuard" in public facing docs.

If we need something more specific than "mapping" we can call if "deobfuscation mapping"

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: It is hardcoded in `create_mapping_upload`

README.md
evkhramkov: fixed, thanks

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: switched to simple version "Path to your Android mapping.txt"

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -295,10 +295,10 @@ def self.available_options
                                   optional: true,
                                       type: String,
                               verify_block: proc do |value|
-                                accepted_formats = ["mapping.txt"]
+                                accepted_formats = [".txt"]
                                 if value
                                   UI.user_error!("Couldn't find mapping.txt at path '#{value}'") unless File.exist?(value)
-                                  UI.user_error!("Only \"mapping.txt\" file name is allowed, you provided \"#{File.name(value)}\"") unless accepted_formats.include? File.basename(value)
+                                  UI.user_error!("Only \"mapping.txt\" file name is allowed, you provided \"#{File.name(value)}\"") unless accepted_formats.include? File.extname(value)
```
bryansmi: If we're only restricting to `.txt`, should we update the language in the error messages to reflect that?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -57,11 +58,38 @@ def self.run_dsym_upload(params)
...
           end
         end
       end
 
+      def self.run_mapping_upload(params)
+        values = params.values
+        api_token = params[:api_token]
+        owner_name = params[:owner_name]
+        app_name = params[:app_name]
+        file = params[:apk]
```
DennisPan: Do we need this?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -158,11 +186,13 @@ def self.get_or_create_app(params)
       def self.run(params)
         values = params.values
         upload_dsym_only = params[:upload_dsym_only]
+        upload_mapping_only = params[:upload_mapping_only]
 
         # if app found or successfully created
         if self.get_or_create_app(params)
-          self.run_release_upload(params) unless upload_dsym_only
-          self.run_dsym_upload(params)
+          self.run_release_upload(params) unless upload_dsym_only unless upload_mapping_only
```
DennisPan: I didn't know you can do two `unless` like this :)

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -158,11 +186,13 @@ def self.get_or_create_app(params)
...
+          self.run_dsym_upload(params)
```
DennisPan: Shouldn't this be skipped if `upload_mapping_only`?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -158,11 +186,13 @@ def self.get_or_create_app(params)
...
         upload_dsym_only = params[:upload_dsym_only]
+        upload_mapping_only = params[:upload_mapping_only]
 
         # if app found or successfully created
         if self.get_or_create_app(params)
-          self.run_release_upload(params) unless upload_dsym_only
-          self.run_dsym_upload(params)
+          self.run_release_upload(params) unless upload_dsym_only unless upload_mapping_only
+          self.run_dsym_upload(params) 
+          self.run_mapping_upload(params)
```
DennisPan: Similar here. Shouldn't this be skipped if `upload_dsym_only`?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: Yep, I missed that, thank you

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -258,6 +290,26 @@ def self.available_options
...
+          FastlaneCore::ConfigItem.new(key: :mapping,
+                                  env_name: "APPCENTER_DISTRIBUTE_ANDROID_MAPPING",
+                               description: "Path to your Android mapping.txt",
+                                  optional: true,
+                                      type: String,
+                              verify_block: proc do |value|
+                                accepted_formats = [".txt"]
+                                if value
+                                  UI.user_error!("Couldn't find mapping.txt at path '#{value}'") unless File.exist?(value)
+                                  UI.user_error!("Only \"mapping.txt\" file name is allowed, you provided \"#{File.name(value)}\"") unless accepted_formats.include? File.extname(value)
```
DennisPan: The error message is confusing. Anything that ends with `.txt` is allowed, right?

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: We don't need `file`, but all the rest is required

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
derpixeldan: For the first error I'd propose to neutralise with "Couldn't find mapping file at path". For the second it depends on what we do according to @ranterle comment.

evkhramkov: Turned out I didn't push that change. Already fixed messages. Second one is `Only \"*.txt\" formats are allowed, you provided ...` to be unified with other error messages

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: yep, make sense

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
evkhramkov: Actually I think it is better to use `unless A || B` here for clarity

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -3,6 +3,7 @@ module Helper
     class AppcenterHelper
 
       # create request
+      # TODO: handle dsym vs mapping
```
ranterle: Does this mean there is still functionality missing?

evkhramkov: no, I think it is just old comment

PR #88, opened 2019-06-13
fix undefined local variable or method `owner'

frederickcook:
Receiving error `NameError: [!] undefined local variable or method `owner' for FastlaneCore::Helper::AppcenterHelper:Class` on upload. I believe this should fix it.


mcmar:
@evkhramkov Can you release a fix with this? I'm currently unable to use this plugin

evkhramkov:
@mcmar are you sure you unable to use plugin because of this particular issue? Bug solved in this PR is not critical, because it occurred only in case of request failure. If you're facing it, it means that you have something wrong with the API request (invalid token or app access)

evkhramkov:
@mcmar we just released `1.1.0` version with this fix, feel free to check it

Issue #89, opened 2019-06-19
Error uploading binary: 500


jlorfeo:
Hello everyone,

when I try to upload a binary via fastlane, I get the following error:

Error uploading binary 500: <!DOCTYPE html>
<html lang='en' class="full_height">
  <head>
    <meta charset='utf-8'>
    <title>Server Error</title>
    <meta content="authenticity_token" name="csrf-param" />
    <link href="/assets/application-default.css" media="all" rel="stylesheet" type="text/css" />
    <link href='/favicon.ico' rel='shortcut icon'>
  </head>
  <body class='full_height'>
    <div class='container-fluid full-height full-width'>
      <div class='row-dialog'>
        <div class='centered-vertical'>
          <div class='centered-horizontal'>
            <div class='well'>
              <h3>Ooops…</h3>
              <p class="pad-top">
                We're sorry, but something went wrong. We've been notified about this issue and we'll take a look at it shortly. If you want to provide more information, please file a support ticket.
              </p>
              <p>
                <a href="/" class="btn btn-ha-primary">Home</a>
                <a href="mailto:support@hockeyapp.net" class="btn">Support</a>
              </p>
            </div>
          </div>
        </div>
      </div>
      <div class='clear'></div>
    </div>
    <script src="/assets/application-default.js" type="text/javascript"></script>
  </body>
</html>

My build is still labeled as success though.
Is this App Center related?

m3nux:
I have same problem :(

setoff:
https://status.appcenter.ms
> Identified - We have to perform an unplanned maintenance on the Distribute service with a potential downtime of one hour. As a consequence some APIs including uploading, downloading and retrieving releases information won't work. 

It is related to service itself not a plugin.
But anyway plugin's fault that it is completed successfully when `appcenter_upload` catch HTTP errors (I've seen 500, 503 during this shutdown). Swallowing errors is bad especially when using fastlane on CI.

derpixeldan:
Yes, we've had an unplanned maintenance. Let's see the good in this, the error handling in case of service unavailability needs to be improved and abort with a respective error. Thanks for identifying and reporting.

bvisin:
Seeing the same issue.  Any ETA when it will be resolved?

jlorfeo:
„[Status] Identified: We're currently running a migration that will fix the issue, however, it takes a few more hours until it's done. Again, we apologize for the trouble this might have caused“

[...]

PR #90, opened 2019-06-21
Fail action on release creation/upload exceptions

ranterle:
Before, the lane would have just swallowed this error. It would have printed it, but still treated the call as a success.
Now, it will fail execution on a service error, i.e. when this is caused by a service outage, to signal the lane did not succeed.

This should fix wrong treatment as mentioned in #89 

Added this behavior only to the case where initial creation of the release failed and when updating the release upload to completed failed.
For cases where symbol or mapping upload fail we typically don't want to fail the entire upload, as by then the release would have been created and re-uploading might cause another release, while retroactively uploading symbols or mappings can be done independently.

### DIFF
--- a/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
+++ b/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
@@ -48,7 +48,7 @@ module Fastlane
         when 404
           UI.error("Not found, invalid owner or application name")
           false
-        when 500..
+        when 500...600
           UI.crash!("Internal Service Error, please try again later")
         else
           UI.error("Error #{response.status}: #{response.body}")
@@ -221,7 +221,7 @@ module Fastlane
         when 200...300
           UI.message("DEBUG: #{JSON.pretty_generate(response.body)}\n") if ENV['DEBUG']
           response.body
-        when 500..
+        when 500...600
           UI.crash!("Internal Service Error, please try again later")
         else
           UI.error("Error #{response.status}: #{response.body}")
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -48,6 +48,8 @@ def self.create_release_upload(api_token, owner_name, app_name)
         when 404
           UI.error("Not found, invalid owner or application name")
           false
+        when 500..
```
lumaxis: The endless range was only introduced in Ruby 2.6 (see https://rubyreferences.github.io/rubychanges/2.6.html#endless-range-1 ).
Do we want to force Ruby >=2.6 by using this? 🤔

ranterle: Good find - did not intend that, switched it to a range with an end value.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -48,6 +48,8 @@ def self.create_release_upload(api_token, owner_name, app_name)
         when 404
           UI.error("Not found, invalid owner or application name")
           false
+        when 500...600
```
DennisPan: Why don't we just stop for other error codes? For example when it's 401, shouldn't we also error and consider the command failed?

ranterle: I think that makes sense - in terms of versioning, this would be a breaking change to existing functionality, so I think we might need to raise the minor version then.

ranterle:
Pushed changes which let the upload action fail entirely if it does not succeed in any way of creating the release or uploading it. This is the main purpose of the action and it should signal failure accordingly to any consumer, i.e. in a continuous integration scenario.

This is a breaking change so raised the minor version number.

PR #95, opened 2019-07-11
Implement support for AAB upload to store

jpeg-ms:
This adds support for Android App Bundle (AAB files) to the stores upload logic.

### DIFF
--- a/lib/fastlane/plugin/appcenter/version.rb
+++ b/lib/fastlane/plugin/appcenter/version.rb
@@ -1,5 +1,5 @@
 module Fastlane
   module Appcenter
-    VERSION = "1.1.0"
+    VERSION = "1.1.1"
   end
 end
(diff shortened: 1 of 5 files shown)


lib/fastlane/plugin/appcenter/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fastlane
   module Appcenter
-    VERSION = "1.1.0"
+    VERSION = "1.1.1"
```
evkhramkov: Please make sure you're on recent master - we already have `1.1.1` plugin version.
Also we usually incrementing version as part of release process, so it is not required to add that in this PR

jpeg-ms: Between the moment I started work on this and now the version has been updated :)
Thanks for the notice, I merged master to resolve this.

PR #100, opened 2019-07-25
Improve owner_name description

Noroxs:
When I tried to upload an ipa file to the app center I accidentally used the `name` instead of the `username`. 
This leads to the following error message:
`Error getting app some/App, 403: {"message"=>"Forbidden", "statusCode"=>403, "code"=>"Forbidden"}`

I had to contact the support to solve this problem.
I adjusted the readme file to specify which name the user should use because there are two different user variables for an account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ To get started, first, [obtain an API token](https://appcenter.ms/settings/apito
 ```ruby
 appcenter_upload(
   api_token: "<appcenter token>",
-  owner_name: "<appcenter account name of the owner of the app (user or organization)>",
+  owner_name: "<appcenter account name of the owner of the app (username or organization URL name)>",
   app_name: "<appcenter app name>",
   apk: "<path to android build binary>",
   notify_testers: true # Set to false if you don't want to notify testers of your new release (default: `false`)
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ To get started, first, [obtain an API token](https://appcenter.ms/settings/apito
 ```ruby
 appcenter_upload(
   api_token: "<appcenter token>",
-  owner_name: "<your appcenter account name (username or organization)>",
-  app_name: "<your app name>",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ To get started, first, [obtain an API token](https://appcenter.ms/settings/apito
 ```ruby
 appcenter_upload(
   api_token: "<appcenter token>",
-  owner_name: "<your appcenter account name>",
+  owner_name: "<your appcenter account username>",
```
evkhramkov: It can also take App Center org name, so it would be better to say:
`<your appcenter account name (user or organization)>`

Noroxs: Thank you for this information! I didn't know. I adjusted the text.

antonys1: If an app is created under **your** account and you have invited collaborators with permission to create builds (i.e. the Developer role) then from **their** perspective it should read `owner_name: "<the appcenter account username that created the app>",`

DennisPan: Yeah agree with that. `your` can be misleading here

lumaxis: @antonys1 @Noroxs I made another attempt at improving the description, let me know what you think!

antonys1: @lumaxis Spot on, thanks

Noroxs: @lumaxis I like it, but I think the `username` should be explicitly listed instead of the `name`.

`"<appcenter account name of the owner of the app (username or organization)>"`

The only reason why I created this pull request was that I stumbled upon this naming issue. Or is it possible to use the `name` instead of `username`?

lumaxis: @Noroxs The problem with that is that we unfortunately name this a bit inconsistently.
- In the UI, we call the decorative, free-text name `"Name:"`. In the API requests and responses, this is `display_name`
- In the UI, we call the programmatic, URL-friendly name `"Username:"`. In the API requests and responses, this is `name`.

I took another try at improving this, maybe this is a good compromise for the moment.

Noroxs: Approved 👍

PR #104, opened 2019-08-14
Support auto-creating apps in App Center organization

evkhramkov:
* Added optional parameter `owner_type` with values `user` and `organization`, default - `user`

Screen Shot 2019-08-13 at 10 10 37 PM

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,6 +14,17 @@ lane :test_autocreation do
     app_name: "MyApplication-notfound-2",
     apk: "./fastlane/app-release.apk"
   )
+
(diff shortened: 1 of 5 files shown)


fastlane/Fastfile
```diff
@@ -14,6 +14,17 @@ lane :test_autocreation do
     app_name: "MyApplication-notfound-2",
     apk: "./fastlane/app-release.apk"
   )
+
+  # app should be created for "fastlane-test-org" organisation
```
DennisPan: `organization`?

evkhramkov: yes, thank for catching

PR #110, opened 2019-08-27
Fail fast when trying to distribute .aab to a group.

jerietve:
Error is now shown before attempting to upload.
While there, remove preview from .aab support.
Ignore .idea folder.


jerietve:
Merged through separate PR already.

Issue #109, opened 2019-08-22
Fetch devices action – How would you prefer I publish?


benkane:
I've created a plugin for fetching a combined list of devices for all distribution groups for an iOS app in App Center (useful when combined with `register_devices` and `match`). Right now I have it locally as a plugin. I was going to publish it as a plugin, but I realized since this plugin is generically named "appcenter", maybe you'd like a PR to add it as an additional action provided by this plugin.

Let me know which you'd prefer – I'm happy to publish the plugin separately or make a PR as an action to be maintained as part of this plugin.

benkane:
Closely related to issue #3

bytenik:
@benkane Did you make your own?

benkane:
@bytenik I did. I’ll try to publish it today. Just need to add some documentation and ideally unit tests

bytenik:
That's fantastic news. I'm finally trying to get off of HockeyApp and this was a big blocker.

bytenik:
Hey, sorry to bug again. Did you have a chance to do this? Thanks!

benkane:
Working on it, sorry for the delay! I will post here when I publish it. If it's not today it will be early next week, I'm actively working on the documentation, but can't dedicate 100% of my time to it

jpeg-ms:
Hi @benkane,
I'm really sorry we didn't check this ticket earlier :( 
I understand you are now working on a separate plugin, but if you eventually decide to send a PR our way, I'd be happy to review it and get it merged.

benkane:
Hi @jpeg-ms, good timing - I was going to publish today :) I have finished unit tests but I'll need to update the documentation now that it'll be under this plugin instead of a separate one. I'll make a PR either today or tomorrow.

cc @bytenik since I know you're waiting for this

benkane:
Added #123

benkane:
Hi @jpeg-ms, I've added the PR for this feature if you want to take a look :)

benkane:
This is now supported, see #123 and #126

PR #123, opened 2019-10-07
Add appcenter_fetch_devices action

benkane:
See #109

This PR adds an action for fetching a list of iOS devices to distribute an app to. It's especially useful in an iOS build pipeline for automatic provisioning of test devices.


lib/fastlane/plugin/appcenter/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fastlane
   module Appcenter
-    VERSION = "1.4.0"
+    VERSION = "1.5.0"
```
benkane: Do I bump this now? I can remove this commit if needed

bytenik:
Looks great! Hopefully MS will merge it shortly!

lib/fastlane/plugin/appcenter/version.rb
phillipleblanc: We will bump it as part of the release process.

phillipleblanc:
Thanks for this PR! This is really cool. I'd like @jpeg-ms or @evkhramkov to take a look at this before we merge this and release it.

lib/fastlane/plugin/appcenter/version.rb
benkane: 👍 I'll remove this

benkane:
Force push was to remove the version number bump commit

jpeg-ms:
Hey this is really cool, but I have a few comments:
- There isn't much in the way of explicit error handling, so we're assuming the errors will be understandable by users (eg. app not found, etc...)
- Shouldn't we use a `destinations` parameter, rather than getting all devices from all groups? We may not want to register all devices of all groups. Maybe '*' or no explicit destination would have the behavior as you've introduced
- I don't think we should require writing to a file. There's likely a cleaner way using shared variables, if documented nicely. But the `devices_file` way is great and should be supported.

benkane:
Thanks for taking a look @jpeg-ms. Some followup comments and questions for you:

> There isn't much in the way of explicit error handling, so we're assuming the errors will be understandable by users (eg. app not found, etc...)

Are you referring to errors that would come back from the API? I can add error handling for those.

> Shouldn't we use a destinations parameter, rather than getting all devices from all groups? We may not want to register all devices of all groups. Maybe '*' or no explicit destination would have the behavior as you've introduced

Maybe? I can't picture a scenario where you wouldn't want every device in your distribution groups for the app to be registered, otherwise some people wouldn't be able to run the app. I don't think a user would set up a distribution group for an app if they were never going to use it. If they use it at all they'd want those devices registered.

I think adding a destinations parameter would be better as a followup as it adds very little to this PR, but this PR as it stands adds usefulness and people are requesting the action. Thoughts?

> I don't think we should require writing to a file. There's likely a cleaner way using shared variables, if documented nicely. But the devices_file way is great and should be supported.
I agree with this since `register_devices` can take a hash of devices. I think using SharedValues might be tough since we'd need to coordinate with a PR to update the `register_devices` action as well, but it'd be nice.

How would you feel about that also being a followup PR? It feels like something that'd be nice to have, but will delay shipping something that already works.

bytenik:
Just to add some other developer color, my team at BallStreet is blocked from switching from HockeyApp to appcenter.ms until either this is merged, released separately, or I write a custom hack for it. So delaying shipping for file creation, etc seems like a bad idea.

The only scenario I can imagine for multiple distribution groups is if you have more than 100 users, and you need to segment your ad-hoc releases so that each one is provisioned with a group of 100. That said, the current HockeyApp fastlane can't support that scenario, so I don't think holding release of this for that reason makes sense.

benkane:
@bytenik  Multiple distribution groups is probably common, e.g. the Collaborators, internal devs, internal qa, and a group for a client testers or some variation of that. Not using every group assigned to an app would be a big edge case though.

Side note, you could grab the plugin from our fork while you're waiting for this to be merged to unblock yourself, then go back to pulling from Ruby Gems once this is merged.

benkane:
@jpeg-ms Additional note regarding destinations – in case it's not obvious, this only grabs devices for all groups assigned to the app, not all groups for the account.

jpeg-ms:
Hey @benkane and @bytenik,

I agree with your comments, I also don't want to block the feature just to avoid creating the devices file. This solution works and is fine.

As for the destinations parameter, I believe it is safer to add it now, including support for '*' to target all groups. Some users can have a huge number of devices, and would partition them across certificates and provisioning profiles. If they were to accidentally register 100 devices to a personal developer account, they'd be stuck with that list for a whole year. To speed things up, I'm working on that myself and will merge this PR first. I hope that's OK with you :)

We can have more follow-up PRs to improve error handling and other things.

benkane:
Awesome, thank you!

jpeg-ms:
@benkane, just a quick note. Please don't take it the wrong way, but I'll keep your name as author in the new action (you've definitely authored it!) but not in the `upload` one, as we're not listing every single contributor individually there :)
Thanks again, I'll be working on a release today.

benkane:
@jpeg-ms That's fine by me! It seems like the standard in the fastlane community is to list all contributors as authors, but I'm fine with whatever your policy is. Thanks for merging this!

jpeg-ms:
> @jpeg-ms That's fine by me! It seems like the standard in the fastlane community is to list all contributors as authors, but I'm fine with whatever your policy is. Thanks for merging this!

I actually did not know that. I'm used to other open source projects and could not find specific guidelines for this one. Thanks for letting me know and thanks again for the PR :)

PR #125, opened 2019-10-10
Do not log public download URL

jpeg-ms:
It contains secret tokens, and should not be printed out to logs that
may be stored in a CI/CD environment. Rather than obfuscating the
tokens and render an invalid URL, let's just not print the download
URL at all.

### DIFF
--- a/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
+++ b/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
@@ -16,12 +16,12 @@ module Fastlane
       end
 
       # create request
-      def self.connection(upload_url = false, dsym = false)
+      def self.connection(upload_url = nil, dsym = false)
         require 'faraday'
         require 'faraday_middleware'
 
         options = {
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -342,14 +343,15 @@ def self.add_to_destination(api_token, owner_name, app_name, release_id, destina
...
+
           download_url = release['download_url']
 
           UI.message("DEBUG: received release #{JSON.pretty_generate(release)}") if ENV['DEBUG']
 
           Actions.lane_context[Fastlane::Actions::SharedValues::APPCENTER_DOWNLOAD_LINK] = download_url
           Actions.lane_context[Fastlane::Actions::SharedValues::APPCENTER_BUILD_INFORMATION] = release
 
-          UI.message("Public Download URL: #{download_url}") if download_url
+          UI.message("Release '#{release_id}' (#{release['short_version']}) was successfully distributed'")
```
jerietve: There's an extra quote at the end

PR #127, opened 2019-10-11
Adds the dsa_signature support

Ajaykn21:
As part of the sparkle update support for macOS and windows releases,
adding the support to capture dsa_signature through via AppCenter_upload_action.


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -499,6 +501,12 @@ def self.available_options
                                        description: "Request timeout in seconds",
                                        optional: true,
                                        type: Integer),
+
+          FastlaneCore::ConfigItem.new(key: :dsa_signature,
+                                       env_name: "APPCENTER_DISTRIBUTE_DSA_SIGNATURE",
+                                       description: "dsa_signature of the macOs or windows releases for Sparkle feed",
```
phillipleblanc: macOS*
Windows*

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -320,6 +320,41 @@ def self.update_release(api_token, owner_name, app_name, release_id, release_not
...
+        release_metadata = nil
+        if dsa_signature.to_s != ''
+          release_metadata = {
+            dsa_signature: dsa_signature
+          }
+        end
+
+        if release_metadata.nil?
+          return
+        end
```
jpeg-ms: This feels like a lot of lines for such a small logic. For now we don't have other metadata so there's no need to make the code complex. Why not just do this:
```ruby
return if dsa_signature.to_s == ''

release_metadata = { ... }
```

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -320,6 +320,41 @@ def self.update_release(api_token, owner_name, app_name, release_id, release_not
...
+          req.headers['internal-request-source'] = "fastlane"
+          req.body = {
+            metadata: release_metadata
+          }
+        end
+
+        case response.status
+        when 200...300
+          UI.message("Release Metadata was successfully updated for release '#{release_id}'")
+        when 404
```
jpeg-ms: 401 should also be handled like in other functions

spec/appcenter_upload_spec.rb
```diff
@@ -999,6 +1007,33 @@ def stub_add_to_destination(status, destination_type = "group", mandatory_update
...
+        it "handles dsa_signature" do
+          stub_check_app(200)
+          stub_create_release_upload(200)
+          stub_upload_build(200)
+          stub_update_release_upload(200, 'committed')
+          stub_update_release(200)
+          stub_update_release_metadata(200)
+          stub_get_destination(200)
+          stub_add_to_destination(200)
+          stub_get_release(200)
```
jpeg-ms: Looks like this test fails because the dsym upload calls have not been stub'ed.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
Ajaykn21: Yeah for now, this release_metadata for only dsa_signature..
But yeah let's handle in future, if any new fields comes up.

PR #135, opened 2019-10-22
show usage of owner_type

export-mike:
I've wasted about 1.5 hrs trying different things due to poor error messages on the api when I submit an app to appcenter using this plugin, I'm using an organisation now - previously a user owner_type. so it worked. It wasn't until I searched the issues for this error: `App with name <app_name> not found, create one?` 

Ideally the api should say something useful. or just resolve to an app in an org...

anyway hopefully this updated readme.md will save some people time in the future! maybe force owner_type to be mandatory?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ To get started, first, [obtain an API token](https://appcenter.ms/settings/apito
 
 ```ruby
 appcenter_fetch_devices(
+  owner_type: "user", # Default is user - set to organisation for appcenter organisations
   api_token: "<appcenter token>",
   owner_name: "<appcenter account name of the owner of the app (username or organization URL name)>",
   app_name: "<appcenter app name>",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,6 +25,7 @@ To get started, first, [obtain an API token](https://appcenter.ms/settings/apito
 
 ```ruby
 appcenter_fetch_devices(
+  owner_type: "user", # Default is user - set to organisation for appcenter organisations
```
jpeg-ms: It's `organization` not `organisation`. I would also place this line right after `owner_type`.

jpeg-ms:
Also, note that we can't make `owner_type` mandatory. It would break existing Fastfiles :)

README.md
export-mike: organisation propper english spelling! I'm sorry chap!

export-mike:
> Also, note that we can't make `owner_type` mandatory. It would break existing Fastfiles :)

Yeah I understand that - maybe the api could return a more useful error message at least...

README.md
export-mike: this has been updated

export-mike: this has been updated

Issue #136, opened 2019-10-22
Support for Windows


adityakhatkhate:
Hi,

I am using this plugin to upload to appcenter from our fastlane setup and it works great! I am however unable to find a way to upload to Windows since the extensions supported by appcenter are not supported by the plugin. Can you please provide an example of a windows upload to appcenter using the plugin.

```
app_name: <my_app_name>,
app_os: "Windows",
app_platform: "Unity",
file: zip_path
```

Thanks.

jpeg-ms:
Hi @adityakhatkhate,

We don't support Windows with the fastlane plugin. We only support ios, android and macOS, like fastlane itself. You may have a look at the [App Center CLI](/microsoft/appcenter-cli) instead, if you have a need for automation.

I'm going to leave this issue open as a feature request. But out of curiosity, are you using other features of fastlane for Windows/Unity builds?

adityakhatkhate:
Hmm, Fastlane did support windows out of the box when it was being used with hockeyapp. We used to create a zip and upload that using Fastlane’s hockeyapp plugin. However, now we have no option to upload the zip in the appcenter plugin.

Yep, we utilize Fastlane for our internal build and deploy processes.

Best,
Aditya.

adityakhatkhate:
Hmm, Fastlane did support windows out of the box when it was being used with hockeyapp. We used to create a zip and upload that using Fastlane’s hockeyapp plugin. However, now we have no option to upload the zip in the appcenter plugin.

jpeg-ms:
Okay, this is great information. I'll see what we can do.

adityakhatkhate:
Thanks Jean-Philippe. Appreciate you looking into this. Our team has been very excited for this move to appcenter and will really appreciate anything you can do to help us. 

From the initial looks of it, it seems like an extension whitelisting error when i try to pass `app_os` as `windows` and `app_platform` as `unity` - ```Only [".aab", ".app", ".app.zip", ".apk", ".dmg", ".ipa", ".pkg"] formats are allowed, you provided ".zip"```

adityakhatkhate:
Hi @jpeg-ms , wanted to follow up on this one to see if it can be made part of the 2019 roadmap before the cutover from hockeyapp to appcenter. This will help us avoid implement an adaptation of fastlane and appcenter cli across the large number of hockeyapp rinks we have.

jpeg-ms:
Hi @adityakhatkhate
[...]

PR #139, opened 2019-10-27
Allow all supported platforms & extensions

jpeg-ms:
This adds explicit support for all known platforms and file extensions currently supported by App Center.
Alternatively, we could add an option to bypass checks and let the API itself accept or reject requests.

This is meant to address issue #136


jerietve:
In my opinion the copying of this logic is getting out of hand. It's in the portal, in the releases service in multiple places, in the CLI, in FastLane. Perhaps we're complicating things too much just to provide errors without roundtrips.

README.md
```diff
@@ -83,15 +83,12 @@ Here is the list of all existing parameters:
...
+| `owner_name` <br/> `APPCENTER_OWNER_NAME` | Owner name as found in the App's URL in App Center |
 | `app_name` <br/> `APPCENTER_APP_NAME` | App name as found in the App's URL in App Center, if there is no app with such name, you will be prompted to create one |
 | `app_display_name` <br/> `APPCENTER_APP_DISPLAY_NAME` | App display name to use when creating a new app |
 | `app_os` <br/> `APPCENTER_APP_OS` | App OS. Used for new app creation, if app 'app_name' was not found |
 | `app_platform` <br/> `APPCENTER_APP_PLATFORM` | App Platform. Used for new app creation, if app 'app_name' was not found |
-| `apk` <br/> `APPCENTER_DISTRIBUTE_APK` | Build release path for android build |
-| `aab` <br/> `APPCENTER_DISTRIBUTE_AAB` | Build release path for android app bundle build |
-| `ipa` <br/> `APPCENTER_DISTRIBUTE_IPA` | Build release path for iOS builds |
-| `file` <br/> `APPCENTER_DISTRIBUTE_FILE` | Build release path for generic builds (.aab, .app, .app.zip, .apk, .dmg, .ipa, .pkg) |
+| `file` <br/> `APPCENTER_DISTRIBUTE_FILE` | Build release path for generic builds |
```
jerietve: 'for generic builds' can now go

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -2,6 +2,18 @@ module Fastlane
   module Actions
     module Constants
       MAX_RELEASE_NOTES_LENGTH = 5000
+      SUPPORTED_EXTENSIONS = {
+          android: %w(.aab .apk),
+          ios: %w(.ipa),
+          mac: %w(.app .app.zip .dmg .pkg),
```
jerietve: The portal does not have .app on the support list. Whether it is required to be ".app.zip" or any .zip is fine also depends on the implementation of this code; it's handled differently in different places.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -131,15 +143,18 @@ def self.run_release_upload(params)
...
-          UI.user_error!("Can't distribute #{file_ext} to stores, please use `destination_type: 'group'`") if %w(.app .app.zip .dmg .pkg).include? file_ext
+          UI.user_error!("Can't distribute #{file_ext} to stores, please use `destination_type: 'group'`") unless Constants::STORE_SUPPORTED_EXTENSIONS.include? file_ext
         end
 
         unless params[:file].to_s.empty?
-          if %w[.dmg .pkg].include? file_ext
+          if Constants::FULL_VERSION_REQUIRED_EXTENSIONS.include? file_ext
             UI.user_error!("Fields `version` and `build_number` must be specified to upload a #{file_ext} file") if build_number.to_s.empty? || version.to_s.empty?
             release_upload_body = { build_version: version, build_number: build_number }
+          elsif Constants::VERSION_REQUIRED_EXTENSIONS.include? file_ext
```
jerietve: file_ext will probably contain ".zip" for ".app.zip" files. This requirement of having a version number should be OS dependent as well.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -297,7 +313,7 @@ def self.available_options
 
           FastlaneCore::ConfigItem.new(key: :app_name,
                                   env_name: "APPCENTER_APP_NAME",
-                               description: "App name as found in the App's URL in App Center. If there is no app with such name, you will be prompted to create one",
+                               description: "App name as found in the App's URL in App Center, if there is no app with such name, you will be prompted to create one",
```
jerietve: The period looks more grammatically correct.

README.md
jpeg-ms: Oops I forgot to re-update the doc after changing in the code

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jpeg-ms: The portal supports .app and automatically zips for you.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jpeg-ms: Good catch!

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -554,7 +578,7 @@ def self.example_code
             api_token: "...",
             owner_name: "appcenter_owner",
             app_name: "testing_app",
-            aab: "./app.aab",
+            file: "./app.aab",
             destinations: "Alpha",
             destination_type: "store",
             build_number: "3",
```
jpeg-ms: Looking at this now I can see that the examples don't make much sense. aab doesn't need build_number, nor does apk.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jpeg-ms: Right, this was meant to unify README and this (they should be copy-paste)

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jpeg-ms: Nevermind my comment. macOS does this for you (at least Chrome on macOS).
But anyway the plugin does this automatic zipping for you :)

Ajaykn21:
> In my opinion the copying of this logic is getting out of hand. It's in the portal, in the releases service in multiple places, in the CLI, in FastLane. 

I feel the same, and for instance, we don't allow .aab for all the stores as well (for Android-Intune, this should throw an error).
And In future, we should be able to handle only at Portal and API level for all these validations, so that other clients like cli, fast lane, etc, just depends on API.

Whats your thoughts on this ? @jpeg-ms

jpeg-ms:
> > In my opinion the copying of this logic is getting out of hand. It's in the portal, in the releases service in multiple places, in the CLI, in FastLane.
> 
> I feel the same, and for instance, we don't allow .aab for all the stores as well (for Android-Intune, this should throw an error).
> And In future, we should be able to handle only at Portal and API level for all these validations, so that other clients like cli, fastlane, etc, just depends on API.
> 
> Whats your thoughts on this ? @jpeg-ms

Yeah I also tend to agree with this. See my initial comment:
> Alternatively, we could add an option to bypass checks and let the API itself accept or reject requests.
We could modify this PR to show warnings only, or refactor the entire logic to ignore these filters and try to do API calls without any initial check.

README.md
```diff
@@ -83,15 +83,12 @@ Here is the list of all existing parameters:
...
+| `owner_name` <br/> `APPCENTER_OWNER_NAME` | Owner name as found in the App's URL in App Center |
+| `app_name` <br/> `APPCENTER_APP_NAME` | App name as found in the App's URL in App Center. If there is no app with such name, you will be prompted to create one |
 | `app_display_name` <br/> `APPCENTER_APP_DISPLAY_NAME` | App display name to use when creating a new app |
 | `app_os` <br/> `APPCENTER_APP_OS` | App OS. Used for new app creation, if app 'app_name' was not found |
 | `app_platform` <br/> `APPCENTER_APP_PLATFORM` | App Platform. Used for new app creation, if app 'app_name' was not found |
-| `apk` <br/> `APPCENTER_DISTRIBUTE_APK` | Build release path for android build |
-| `aab` <br/> `APPCENTER_DISTRIBUTE_AAB` | Build release path for android app bundle build |
-| `ipa` <br/> `APPCENTER_DISTRIBUTE_IPA` | Build release path for iOS builds |
-| `file` <br/> `APPCENTER_DISTRIBUTE_FILE` | Build release path for generic builds (.aab, .app, .app.zip, .apk, .dmg, .ipa, .pkg) |
+| `file` <br/> `APPCENTER_DISTRIBUTE_FILE` | File path to the release build to publish |
```
Ajaykn21: This is cool..

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -336,52 +357,59 @@ def self.available_options
                               verify_block: proc do |value|
                                 accepted_formats = [".apk"]
                                 file_extname_full = Helper::AppcenterHelper.file_extname_full(value)
-                                UI.user_error!("Only \".apk\" formats are allowed, you provided \"#{file_extname_full}\"") unless accepted_formats.include? file_extname_full
+                                self.optional_error("Only \".apk\" formats are allowed, you provided \"#{file_extname_full}\"") unless accepted_formats.include? file_extname_full
                               end),
 
           FastlaneCore::ConfigItem.new(key: :aab,
```
Ajaykn21: I assume we are keeping this now for backward compatibility ?
Whats the plan to remove this ?

jpeg-ms: No plan to remove. Just marked as deprecated. :)

jpeg-ms: To be clear, to remove we would have to change the major version number

PR #140, opened 2019-10-28
Use existing timeout option for dSYM uploads

### LINKED ISSUE #137: Add timeout usage to DSYM uploads too.
The IPA upload respects the timeout option from #91 ; however the timeout option should also be used for the _any_ upload like the DSYMs.

I often see successful uploads of the IPA only to fail on DSYM upload.

adingler6:
Closes #137

I thought about adding the timeout as the default in the connection builder, but I thought it would be best to only use it when needed.


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -501,7 +503,7 @@ def self.available_options
 
           FastlaneCore::ConfigItem.new(key: :timeout,
                                        env_name: "APPCENTER_DISTRIBUTE_TIMEOUT",
-                                       description: "Request timeout in seconds",
+                                       description: "Request timeout in seconds. Used when uploading builds and dSYM files",
```
jpeg-ms: dSYM or mapping

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -163,7 +165,7 @@ def self.update_symbol_upload(api_token, owner_name, app_name, symbol_upload_id,
       # upload symbol (dSYM or mapping) files to specified upload url
       # if succeed, then commits the upload
       # otherwise aborts
-      def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, symbol_upload_id, upload_url)
+      def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, symbol_upload_id, upload_url, timeout)
         connection = self.connection(upload_url, true)
 
         response = connection.put do |req|
```
jpeg-ms: This is the place where the real file upload happens, as you can see from the file IO. This is where the timeout option should be added to the request.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -50,15 +51,15 @@ def self.run_dsym_upload(params)
             file_name = File.basename(dsym_path)
             dsym_upload_details = Helper::AppcenterHelper.create_mapping_upload(api_token, owner_name, app_name, file_name ,build_number, version)
           else
-            dsym_upload_details = Helper::AppcenterHelper.create_dsym_upload(api_token, owner_name, app_name)
+            dsym_upload_details = Helper::AppcenterHelper.create_dsym_upload(api_token, owner_name, app_name, timeout)
```
jpeg-ms: This only creates an upload URL & ID, so timeout due to large files is not going to happen here.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -136,10 +137,11 @@ def self.create_dsym_upload(api_token, owner_name, app_name)
       end
 
       # commits or aborts symbol upload
-      def self.update_symbol_upload(api_token, owner_name, app_name, symbol_upload_id, status)
+      def self.update_symbol_upload(api_token, owner_name, app_name, symbol_upload_id, status, timeout)
```
jpeg-ms: On the other hand, this is not uploading anything, merely doing a quick PATCH call to the API to signal that the upload has completed. So this timeout is not necessary.

adingler6:
@jpeg-ms Thanks for the review! Your feedback made sense and I updated.

jpeg-ms:
Hey so I'm trying to test this PR and my uploads never time out, including IPA uploads. It's as if Faraday does not respect the `timeout` option at all. Does this actually work for you?

adingler6:
So I just did some tests. It does work on IPA uploads, but your right, I doesn't seem to respect it on the dSYM upload. Very strange.

This is what I did to test the IPA...
First I dramatically increased the size of `./fastlane/app-release.ipa` and ran:
```
appcenter_upload(
    api_token: ENV["TEST_APPCENTER_API_TOKEN"],
    owner_name: ENV["TEST_APPCENTER_OWNER_NAME"],
    app_name: "MyApplication-01",
    ipa: "./fastlane/app-release.ipa",
    destinations: ENV["TEST_APPCENTER_DISTRIBUTE_GROUP"],
    destination_type: "group",
    release_notes: ENV["TEST_APPCENTER_DISTRIBUTE_RELEASE_NOTES"],
    timeout: 1
  )
```

This gave me a timeout error `Faraday::TimeoutError: [!] Net::ReadTimeout`. Then If I change the timeout to `10` it works and uploads to AppCenter.

I tried the same thing with the dSYM zip to see if it would timeout and it did not. I then tried to see if I could get it to hit the 60 seconds default timeout, but after almost 2 mintues I got this error
` Error uploading dSYM 413: ﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>RequestBodyTooLarge</Code><Message>The request body is too large and exceeds the maximum permissible limit.`

@jaleksynas, what was the error you were getting that led to the issue being created? Hard for me to reproduce the timeout.

jpeg-ms:
So it seems that Faraday does not respect the timeout when executing a PUT. I've seen elsewhere[1] that it does not respect timeout on GET when using this approach.

For dSYM I tried twice and had both times a timeout at 2 minutes :(

[1] https://stackoverflow.com/questions/32272374/faraday-timeout-on-a-simple-get

jaleksynas:
Any reason this api is only PUT? POST works, can a post be used instead of
put?

jpeg-ms:
> Any reason this api is only PUT? POST works, can a post be used instead of put?

Yes, indeed. We are using the PUT API defined here: https://docs.microsoft.com/en-us/rest/api/storageservices/put-block

I think it's better to investigate alternative solutions than use an unsupported POST call. Either we find a way to make Faraday respect this timeout, or we'll have to switch to a better library (`http.rb` comes to mind).

jpeg-ms:
Another thing we may want to investigate is optimize the upload by zipping the file. I think Azure blobs handle zip very nicely.

jaleksynas:
I think DSYMs are already zipped.  Faraday uses `http.rb` under the covers (the default adapter) and it seems that PUT doesn't support a timeout like GET and POST do.  There are other middle-ware options for Faraday but those would require significant investigation.

jp-andre:
I'm closing this as this has been stalled for too long. This approach unfortunately does not work for now. I would love to find a working solution though!

PR #150, opened 2019-11-07
Adding support for * as destinations

Hanstad:
Distributing build to all distribution groups by using the '*' as destination. Make exceptions by adding distributions groups to destination_excludes

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -203,7 +203,7 @@ module Fastlane
 
               UI.abort_with_message!("Failed to list distribution groups for #{owner_name}/#{app_name}") unless distribution_groups
               
-              destination_array = distribution_groups.map {|h| h['name'].downcase }
+              destination_array = distribution_groups.map {|h| h['name'] }
             else
               destinations_array = destinations.split(',').map(&:strip)
             end
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -213,7 +213,7 @@ module Fastlane
                 UI.important("Could not find the excluded distribution group: #{group}") unless distribution_group_names.include? group.downcase 
               end
             else
-              destinations_array = destinations.split(',')
+              destinations_array = destinations.split(',').map(&:strip)
             end
             
             destinations_array.each do |destination_name|
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -108,6 +108,7 @@ module Fastlane
         owner_type = params[:owner_type]
         app_name = params[:app_name]
         destinations = params[:destinations]
+        destination_excludes = params[:destination_excludes]
         destination_type = params[:destination_type]
         mandatory_update = params[:mandatory_update]
         notify_testers = params[:notify_testers]
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -108,7 +108,6 @@ module Fastlane
         owner_type = params[:owner_type]
         app_name = params[:app_name]
         destinations = params[:destinations]
-        destination_excludes = params[:destination_excludes]
         destination_type = params[:destination_type]
         mandatory_update = params[:mandatory_update]
         notify_testers = params[:notify_testers]
@@ -202,16 +201,9 @@ module Fastlane
                 app_name: app_name
               )
 
-              destination_excludes_array = destination_excludes.split(',').map(&:strip).map(&:downcase)
               UI.abort_with_message!("Failed to list distribution groups for #{owner_name}/#{app_name}") unless distribution_groups
-              distribution_groups.each do |group|
-                destinations_array << group['name'] unless destination_excludes_array.include? group['name'].downcase 
-              end
-
-              distribution_group_names = distribution_groups.map {|h| h['name'].downcase }
-              destination_excludes_array.each do |group|
-                UI.important("Could not find the excluded distribution group: #{group}") unless distribution_group_names.include? group.downcase 
-              end
+              
+              destination_array = distribution_groups.map {|h| h['name'].downcase }
             else
               destinations_array = destinations.split(',').map(&:strip)
             end
@@ -518,13 +510,6 @@ module Fastlane
                                   optional: true,
                                       type: String),
 
-          FastlaneCore::ConfigItem.new(key: :destination_excludes,
-                                  env_name: "APPCENTER_DISTRIBUTE_DESTINATION_EXCLUDES",
-                               description: "Comma separated list of destination names to be excluded, if destinations is '*'",
-                             default_value: "",
-                                  optional: true,
-                                      type: String),
-
           FastlaneCore::ConfigItem.new(key: :destination_type,
                                   env_name: "APPCENTER_DISTRIBUTE_DESTINATION_TYPE",
                                description: "Destination type of distribution destination. 'group' and 'store' are supported",
@@ -644,7 +629,6 @@ module Fastlane
             app_name: "testing_ios_app",
             file: "./app-release.ipa",
             destinations: "*",
-            destination_excludes: "Collaborators",
             destination_type: "group",
             release_notes: "release notes",
             notify_testers: false
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -505,7 +505,7 @@ module Fastlane
 
           FastlaneCore::ConfigItem.new(key: :destinations,
                                   env_name: "APPCENTER_DISTRIBUTE_DESTINATIONS",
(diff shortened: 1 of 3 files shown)


phillipleblanc:
This isn't very clear that the '*' only applies to distribution groups. We also support distributing to Store destinations and testers directly.

Hanstad:
> This isn't very clear that the '*' only applies to distribution groups. We also support distributing to Store destinations and testers directly.

What do you suggest that we do? Will it be possible to support Store as well? Or should we introduce yet another option?

Hanstad:
@b3hni4

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -596,7 +620,17 @@ def self.example_code
...
+          )',
+          'appcenter_upload(
+            api_token: "...",
+            owner_name: "appcenter_owner",
+            app_name: "testing_google_play_app",
+            file: "./app.aab",
+            destinations: "*",
+            destination_excludes: "Collaborators",
+            destination_type: "store",
+            release_notes: "this is a store release"
```
phillipleblanc: This example is wrong - it would be incorrect to have a destination as "*" for a store release

phillipleblanc:
> > This isn't very clear that the '*' only applies to distribution groups. We also support distributing to Store destinations and testers directly.
> 
> What do you suggest that we do? Will it be possible to support Store as well? Or should we introduce yet another option?

I can't imagine a scenario in which anyone would want to distribute to all Store destinations, since that includes Production as well as any beta testing groups supported by the Stores.

Hanstad:
I could change the description of the option to state that this only works for distribution groups. How does that sound? If it does not make sense for Store lists, but does make sense for Distrubution lists, then that is the only option unless you will go splitting the option into a distribution-list and a store-list and by that removing the backwards compatibility

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Hanstad: I agree. I need to change the destination_type

phillipleblanc:
Another way would be to error out if `*` was specified for the destinations and the `destination_type` wasn't `group`

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -191,7 +192,24 @@ def self.run_release_upload(params)
             release = Helper::AppcenterHelper.update_release(api_token, owner_name, app_name, release_id, release_notes)
             Helper::AppcenterHelper.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature)
 
-            destinations_array = destinations.split(',')
+            destinations_array = []
+            if destinations == '*'
```
jpeg-ms: At this point we must make sure that the distribution type is `group`.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jpeg-ms: And also `.aab` is only valid for google play :)
The example Fastfile should also be modified to show off this feature.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -108,6 +108,7 @@ def self.run_release_upload(params)
         owner_type = params[:owner_type]
         app_name = params[:app_name]
         destinations = params[:destinations]
+        destination_excludes = params[:destination_excludes]
```
Ajaykn21: Hi @Hanstad 
Agree with @jpeg-ms, when we can achieve it with existing options itself.
Please check if you can remove `destination_excludes` .

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Hanstad: Will add this :)

Hanstad:
I have implemented an error-out if destinations are set to '*' while destination_type is not 'group'.
Furthermore I have implemented a Important message if the destination_excludes contains groups that are not found on AppCenter.

Hanstad:
I would appreciate if someone could give feedback or merge my PR

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -191,7 +193,29 @@ def self.run_release_upload(params)
...
+              distribution_groups.each do |group|
+                destinations_array << group['name'] unless destination_excludes_array.include? group['name'].downcase 
+              end
+
+              distribution_group_names = distribution_groups.map {|h| h['name'].downcase }
+              destination_excludes_array.each do |group|
+                UI.important("Could not find the excluded distribution group: #{group}") unless distribution_group_names.include? group.downcase 
+              end
+            else
+              destinations_array = destinations.split(',')
```
jpeg-ms: It might be worth it adding a `map(&:strip)` here as well while we're at it.

Hanstad:
Thanks for your questions. 

Use case: 
We have a app that is build using fastlane on different branches (development, rc, master etc). Each build is uploaded to is own app on AppCenter (app_master, app_development etc.) but uses the same Fastfile.
Depending on the app on AppCenter (app_master, app_development etc.) different groups of people receives the build. This list changes all the time. 
We would like to control who gets the build on AppCenter by adding or removing distribution groups and not change the fastlane code everytime. 

I'm not a fan of the exclude either, but the list received from `fetch_distribution_groups` contains the **Collaborators** group, which I don't want to distribute to. This is a group that I cannot remove (AFAIK). So I introduced the exclude to remove this group. An alternative would be to hardcode the removal of **Collaborators**

Please let me know which approach you prefer.

I will start working on unit tests for this.

jpeg-ms:
Interesting, thanks the detailed response.
It sounds to me like you could use the environment variable `APPCENTER_DISTRIBUTE_DESTINATIONS` instead of modifying the Fastfile, to select which groups to send the release to.

Again out of curiosity, why not distribute to Collaborators? As owners of the app, shouldn't they get notified when there's a new version?

Hanstad:
Thanks for your proposal. The number of groups changes all the time, so if I used the env. variable I would have to update the Jenkinsfile (we are using Jenkins) all the time (equal to update the Fastfile as it will require a commit) or update all our build agents all the time. So I don't think that is a good solution.

Collaborators are all people with interest of the app but not all of these needs the build. Some looks at diagnostics and some looks at analytics only.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Hanstad: map(&:strip) added as well as unit tests

jpeg-ms:
Hey @hanstad

I'm sorry to say this but we aren't too excited about the excludes part. Adding support for `'*'` seems OK to us, but excludes seems like a non-necessary feature. It will not be clear how that works with standard `destinations`, and that doesn't seem like a feature we want to have to maintain in this plugin.
Note that in your case, Collaborators all have access to the app and all releases anyway, so it's not a security feature. Users who don't want to receive notifications about new releases can opt out from email notifications themselves.

Alternatively you could look into using the CLI to list the groups prior to calling fastlane, or even distribute using the CLI.

Hanstad:
Hey @jpeg-ms 
I'm sad about your decision, but that is sometimes how it is when using 3'rd party plugins. I will update the PR to exclude the exclude option. Are there anyway that I can introduce a "secret" env variable for the excludes instead?

jpeg-ms:
If we introduce something that isn't documented (secret as you say), then we will end up not maintaining it, which would only make it worse for all of us on the longer term.

Hanstad:
I have now removed the support for destination_excludes

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -202,16 +201,9 @@ def self.run_release_upload(params)
...
-              distribution_groups.each do |group|
-                destinations_array << group['name'] unless destination_excludes_array.include? group['name'].downcase 
-              end
-
-              distribution_group_names = distribution_groups.map {|h| h['name'].downcase }
-              destination_excludes_array.each do |group|
-                UI.important("Could not find the excluded distribution group: #{group}") unless distribution_group_names.include? group.downcase 
-              end
+              
+              destination_array = distribution_groups.map {|h| h['name'].downcase }
```
jpeg-ms: I don't think we should downcase here (it apparently works but I wouldn't count on it)

Hanstad: I agree, so I have removed it

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -489,12 +505,11 @@ def self.available_options
 
           FastlaneCore::ConfigItem.new(key: :destinations,
                                   env_name: "APPCENTER_DISTRIBUTE_DESTINATIONS",
-                               description: "Comma separated list of destination names. Both distribution groups and stores are supported. All names are required to be of the same destination type",
+                               description: "Comma separated list of destination names, use '*' for all distribution groups and exclude groups with destination_excludes. Both distribution groups and stores are supported. All names are required to be of the same destination type",
```
jpeg-ms: Oops :) Need to update comment

Hanstad: Great catch!

Hanstad:
You are welcome and thanks for your feedback even though it was not what I wished for :)

Issue #154, opened 2019-11-12
No install button for apps released via fastlane plugin


tommeier:
Hi,

We've been using HockeyApp for many years - with fastlane plugin for uploading + distributing. 
I've recently migrated all apps to Microsoft App Center - (given the 16th November deadline).

However, I've upgraded to the App Center SDK - and after generating a release - the email includes an 'Install' button - but when clicking on that it shows a screen displaying "The developers are working on a version for your device". When clicking the previous versions - the 'Install' button is only available on the Hockey old versions but not the new App Center versions.

So all our _old_ migrated releases (generated via fastlane + hockeyapp) are available to install - Install button is visible. But all _new_ apps released with fastlane + `fastlane-plugin-appcenter` never has an 'install' button on the page.

---

I've narrowed this purely to fastlane plugin because if I manually release the same `.ipa` generated via fastlane - the install button is visible for devices in the distribution groups - all settings AFAIK are the same.

Our testing has been completed around here: https://appcenter.ms/orgs/ferocia-pty-ltd/apps/Test/distribute/releases

There is no visible reason that I can see, or diagnose with, from the UI why the fastlane distributed apps have no install button.

Example of successful upload, now with strict mode, but no install button for distribution groups:

```
[14:43:04]: Cruising over to lane 'ios release_to_app_center' 🚖
[14:43:04]: Uploading to Microsoft App Center with options:
[14:43:04]: {:owner_name=>"ferocia-pty-ltd", :owner_type=>"organization", :notify_testers=>true, :release_notes=>"## Release: v4.6.36 (305) - Tuesday, 12 Nov 2019 2:39 PM\n\nBumped release with Microsoft App Center\n\nLatest tweaks to App Center deploy via strict mode uploads\n", :strict=>"true", :destinations=>"Administrators-Distribution-Group,x-Test-Team,x-Distribution-Group", :file=>"/Users/canary/src/x-ios/x-x-x-Test.ipa", :dsym=>"/Users/canary/src/x-ios/x-x-x-Test.app.dSYM.zip", :app_name=>"Test", :api_token=>"xxx"}
[14:43:04]: ------------------------------
[14:43:04]: --- Step: appcenter_upload ---
[14:43:04]: ------------------------------
[14:43:04]: Enabled strict mode
[14:43:05]: Starting release upload...
[14:43:06]: Uploading release binary...
[14:43:17]: Binary uploaded
[...]

Issue #28, opened 2018-01-08
Unable to upload Android APK after uploading iOS IPA


wmcbain:
I'm working with a React Native project and what we're trying to achieve is building iOS and Android in the same build. However I'm unable to upload the Android APK after uploading the iOS IPA with the following error:

```
ERROR [2018-01-08 11:49:38.79]: Couldn't find dSYM file at path '/Users/vagrant/git/App.app.dSYM.zip'
ERROR [2018-01-08 11:49:38.79]: You passed invalid parameters to 'appcenter_upload'.
ERROR [2018-01-08 11:49:38.79]: Check out the error below and available options by running `fastlane action appcenter_upload`
```

I can't share my entire Fastfile without some editing but in the ios lane we run

```
appcenter_upload(
  api_token: "token",
  owner_name: "Owner",
  app_name: "App",
  ipa: binary_path
)
```

Followed by uploading dsyms after they're downloaded:

```
dsyms = lane_context[SharedValues::DSYM_PATHS]
dsyms.each do |dsym|
  appcenter_upload(
    api_token: "token",
    owner_name: "Owner",
    app_name: "App",
    upload_dsym_only: true,
    dsym: dsym
  )
end
```

After this completes we build Android and in that lane run:

```
appcenter_upload(
  api_token: "Token",
  owner_name: "Owner",
  app_name: "App",
  apk: "./android/app/build/outputs/apk/app-releaseStaging.apk"
)
```

`:APPCENTER_BUILD_INFORMATION` at the point of Android is set to the iOS project unsure if it helps, but thought that it might. Thanks!

evkhramkov:
@wmcbain does it work if you build and upload only Android?

wmcbain:
@evkhramkov It does work when Android is run on it's own. Do I need to clear env variables in between steps?

richardmtsr:
@evkhramkov Is there an update on this? I'm having the same issue. Really want to be able to run both builds together but at the moment we are restricted.

amalshah71:
Any update on this ?

If i even do only Android, it says :: "couldn't find build path..."

evkhramkov:
@richardmtsr @amalshah71 @wmcbain sorry for so late response.
Is this still an issue for you?

[...]

PR #156, opened 2019-11-14
Fix: Priority of deprecated fields and file parameter

tommeier:
Related to: https://github.com/microsoft/fastlane-plugin-appcenter/issues/154

Hi 👋 , took a while to track this one down.

The effect of this bug is not seeing install buttons in Microsoft App Center when released via fastlane but visible when manually creating release via web UI.

We build 2 packages in the one lane for distribution - one via Microsoft App Center (was Hockey) and the other via testflight - 2 streams and 2 different targets.

The first build we send to MS App Center, but only _after_ building the second release, then packaging both with github PRs etc. 

The `ipa` option is deprecated on `appcenter_upload` commands so we use `file`. However there is a mismatch of using `.detect` on the parameters given as that is choosing the _default_ `IPA_OUTPUT_PATH` which has been set by the _second_ Gym build and ignoring the passed `file` parameter.

This PR ensures any passed `file` param is the priority before looking at default values being passed.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -132,10 +132,10 @@ module Fastlane
         end
 
         file = [
+          params[:file],
           params[:ipa],
           params[:apk],
           params[:aab],
-          params[:file]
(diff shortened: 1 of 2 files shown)


jpeg-ms:
Also I believe this will solve #28.

Issue #143, opened 2019-10-31
get_or_create_app error


jenkins-endios:
can't not build under Gem: 3.0.6 and ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin19] ??

44:in `run'
/Library/Ruby/Gems/2.6.0/gems/fastlane-plugin-appcenter-1.5.0/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb:228:in `get_or_create_app': [!] undefined method `length' for nil:NilClass (NoMethodError)

jpeg-ms:
Could please you share your Fastfile? It could be related to a syntax error.

phillipleblanc:
@jenkins-endios - Were you able to figure this out?

Ajaykn21:
@jenkins-endios We are able to repro this issue and working on the fix.

Ajaykn21:
#157 linking the PR

Ajaykn21:
This is part of v1.7.0

PR #157, opened 2019-11-14
Fixing the app creation by handling os correctly

Ajaykn21:
Handling platforms check properly.
Since we verify if the app_platform is provided, if not we prompt the user to select from our `platforms` mapping.
And while doing so, there is an error where we pass platforms[os] instead it should be platforms[os.to_sym]

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -232,7 +232,7 @@ module Fastlane
           iOS: %w[Objective-C-Swift React-Native Xamarin],
           macOS: %w[Objective-C-Swift]
         }
-        
+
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -232,7 +232,7 @@ def self.get_or_create_app(params)
           iOS: %w[Objective-C-Swift React-Native Xamarin],
           macOS: %w[Objective-C-Swift]
         }
-
+
```
jpeg-ms: oops

Ajaykn21: 😬 Done

Ajaykn21:
#143 should be solved by this.

Issue #152, opened 2019-11-11
Error while running appcenter_upload . bad URI(is not URI?)


rkhandelwal17:
App Center has the URL formatted like this:
https://appcenter.ms/users/myname/apps/My-App-Name
But the value generated in fastlane is:

/v0.1/apps/mynamehere/My App Name

<details><summary>✅ fastlane environment ✅</summary>

### Stack

| Key                         | Value                                       |
| --------------------------- | ------------------------------------------- |
| OS                          | 10.14.5                                     |
| Ruby                        | 2.3.8                                       |
| Bundler?                    | false                                       |
| Git                         | git version 2.17.2 (Apple Git-113)          |
| Installation Source         | /usr/local/lib/ruby/gems/2.3.0/bin/fastlane |
| Host                        | Mac OS X 10.14.5 (18F132)                   |
| Ruby Lib Dir                | /usr/local/Cellar/ruby@2.3/2.3.8_1/lib      |
| OpenSSL Version             | OpenSSL 1.0.2q  20 Nov 2018                 |
| Is contained                | false                                       |
| Is homebrew                 | false                                       |
| Is installed via Fabric.app | false                                       |
| Xcode Path                  | /Applications/Xcode.app/Contents/Developer/ |
| Xcode Version               | 10.1                                        |

### System Locale

| Variable | Value       |   |
| -------- | ----------- | - |
| LANG     | en_AU.UTF-8 | ✅ |
| LC_ALL   |             |   |
| LANGUAGE |             |   |

### fastlane files:

<details><summary>`./fastlane/Fastfile`</summary>

```ruby
opt_out_usage

fastlane_version "2.109.1"
xcversion(version: "10.1")

default_platform :ios

# Environment
artifactory_endpoint = "https://artifactory.srv.westpac.com.au/artifactory"
carthage_repo = "A00931_Carthag"
#ipa_path = "Test/Snapshot/au/com/westpac/native/Banking.Universal.iOS/#{last_git_commit[:abbreviated_commit_hash]}/"
#new_framework_path = "#{prebuild_options[:package_path]}/#{new_ver}/
#ipa_path = "#{artifactory_endpoint}/#{carthage_repo}/au/com/westpac/native/Banking.Universal.ios"

type_mapping = {
  "develop" => { folder: "Develop", pre_type: "alpha" },
  "snapshot" => { folder: "Snapshot", pre_type: "alpha" },
[...]

PR #158, opened 2019-11-14
Fixing the platform warning and  bad URI(is not URI?) error

Ajaykn21:
# Issue: 1
As part of the config key :file, we are checking the supported file extensions based on environment variable `SharedValues::PLATFORM_NAME`.
And if this is not present, we are showing incorrect error message.

`[19:34:18]: Unknown platform '', consider using one of: [:android, :ios, :mac, :windows, :custom]`

This is slightly confusing when users are using this for only upload action and doesn't want to set the platform.

# Issue 2
When user gives AppName with spaces/special characters, then url construction for GET is failing with **URI::InvalidURIError: [!] bad URI(is not URI?): v0.1/apps/korea/test app**
If we do URI.encode this error is resolved but fails in app_creation steps with validation errors.

Hence handling this just by rescue-ing on URI::InvalidURIError.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -238,7 +238,7 @@ module Fastlane
             return true
           end
         rescue URI::InvalidURIError
-          UI.user_error!("Provided app_name: '#{app_name}' is not in a valid format. Please ensure no special charecters and spaces in the app_name.")
+          UI.user_error!("Provided app_name: '#{app_name}' is not in a valid format. Please ensure no special characters or spaces in the app_name.")
           return false
         end
 
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -238,7 +238,7 @@ module Fastlane
             return true
           end
         rescue URI::InvalidURIError
-          UI.error("Provided app_name: '#{app_name}' is not in a valid format. Please ensure no special charecters and spaces in the app_name.")
+          UI.user_error!("Provided app_name: '#{app_name}' is not in a valid format. Please ensure no special charecters and spaces in the app_name.")
           return false
         end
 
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -233,8 +233,13 @@ def self.get_or_create_app(params)
...
         }
 
-        if Helper::AppcenterHelper.get_app(api_token, owner_name, app_name)
-          return true
+        begin
+          if Helper::AppcenterHelper.get_app(api_token, owner_name, app_name)
+            return true
+          end
+        rescue URI::InvalidURIError
+          UI.user_error!("Provided app_name: '#{app_name}' is not in a valid format. Please ensure no special charecters and spaces in the app_name.")
```
jpeg-ms: Typo here: "characters or spaces"

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -407,13 +412,15 @@ def self.available_options
...
                               verify_block: proc do |value|
                                 platform = Actions.lane_context[SharedValues::PLATFORM_NAME]
-                                accepted_formats = Constants::SUPPORTED_EXTENSIONS[platform.to_sym] if platform
-                                unless accepted_formats
-                                  UI.important("Unknown platform '#{platform}', consider using one of: #{Constants::SUPPORTED_EXTENSIONS.keys}")
-                                  accepted_formats = Constants::ALL_SUPPORTED_EXTENSIONS
+                                if platform
+                                  accepted_formats = Constants::SUPPORTED_EXTENSIONS[platform.to_sym]
+                                  unless accepted_formats
+                                    UI.important("Unknown platform '#{platform}', Supported are #{Constants::SUPPORTED_EXTENSIONS.keys}")
```
jpeg-ms: Why change the text? Maybe rephrase as "Supported platforms: "

Ajaykn21: Original text suggests user to set which is confusing: `consider using one of:`.. hence changed

jpeg-ms:
This is to fix #152

PR #159, opened 2019-11-15
Adding support to upload_build_only option

Ajaykn21:
This is to support few feature requests where customers just want to upload the builds but not symbols.

Hence introduced a new flag called `upload_build_only` 
And when set to true, this will upload build only and skips symbols/mapping

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -28,7 +28,7 @@ module Fastlane
         api_token = params[:api_token]
         owner_name = params[:owner_name]
         app_name = params[:app_name]
-        file = params[:ipa]
+        file = params[:file] || params[:ipa]
         dsym = params[:dsym]
         build_number = params[:build_number]
         version = params[:version]
@@ -261,6 +261,7 @@ module Fastlane
 
       def self.run(params)
         values = params.values
+        upload_build_only = params[:upload_build_only]
         upload_dsym_only = params[:upload_dsym_only]
         upload_mapping_only = params[:upload_mapping_only]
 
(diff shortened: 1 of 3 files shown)


jpeg-ms:
Related to #137 (but does not address the timeout).

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -422,6 +423,17 @@ def self.available_options
...
+          
+          FastlaneCore::ConfigItem.new(key: :upload_build_only,
+                                  env_name: "APPCENTER_DISTRIBUTE_UPLOAD_BUILD_ONLY",
+                               description: "Flag to upload only the build to App Center. Skips uploading symbols or mapping",
+                                  optional: true,
+                                 is_string: false,
+                             default_value: false,
+                       conflicting_options: [:upload_dsym_only, :upload_mapping_only],
+                            conflict_block: proc do |value|
+                              UI.user_error!("You can't use 'upload_build_only' and '#{value.key}' options in one run")
```
jpeg-ms: Interestingly, this is only true if the flag was set to `true`. If we explicitly set to `false` then the options are somewhat compatible.

spec/appcenter_upload_spec.rb
```diff
@@ -1498,6 +1498,54 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
       expect(values[:dsym_path]).to eq('./spec/fixtures/symbols/Themoji.dSYM.zip')
     end
 
+    it "allows to upload build only even if dsym provided when upload_build_only is true"  do
```
jpeg-ms: spec/appcenter_upload_spec.rb:1501:90: C: Layout/ExtraSpacing: Unnecessary spacing detected.
    it "allows to upload build only even if dsym provided when upload_build_only is true"  do

Ajaykn21: fixed

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Ajaykn21: oh yeah. nice catch. will remove this conflict block.. because we can't check the values set to true/false.

Ajaykn21: As discussed, we are going with this as it won't make sense to provide those fields at same time, even if its true/false

PR #160, opened 2019-11-15
Adding windows support for app creation when not exists

Ajaykn21:
As part of this plugin, we support when app doesn't exists, we have an option to create.
And when we don't provide app_platform, we prompt the users to select from supported platforms.
But this was not supported in Windows case.

Added support to this.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -230,7 +230,8 @@ module Fastlane
         platforms = {
           Android: %w[Java React-Native Xamarin],
           iOS: %w[Objective-C-Swift React-Native Xamarin],
-          macOS: %w[Objective-C-Swift]
+          macOS: %w[Objective-C-Swift],
+          Windows: %w[UWP WPF WinForms Unity]
         }
 
         begin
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -230,7 +230,8 @@ def self.get_or_create_app(params)
         platforms = {
           Android: %w[Java React-Native Xamarin],
           iOS: %w[Objective-C-Swift React-Native Xamarin],
-          macOS: %w[Objective-C-Swift]
+          macOS: %w[Objective-C-Swift],
```
jpeg-ms: How about having Unity for ios and android?
Should we also have tvOS?

spec/appcenter_upload_spec.rb
```diff
@@ -1300,6 +1300,32 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
       end").runner.execute(:test)
     end
 
+    it "creates app when app_os is Windows amd selects the app_platform" do
```
jpeg-ms: and

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Ajaykn21: Not sure, how symbols and stuff works for Unity iOS..
For tvOS, there is no distribute support from portal.
Lets ignore them.. For now this should be fine.

PR #171, opened 2019-12-02
Fix rules for latest rubocop 0.77

jpeg-ms:

PR #172, opened 2019-12-02
Fix use of new constant from fastlane

jpeg-ms:


jpeg-ms:
This fixes issues introduced by #169. My bad for not checking closely in which version of fastlane this new constant was introduced.

.rubocop.yml
```diff
@@ -69,10 +69,10 @@ Lint/UselessAssignment:
     - '**/spec/**/*'
 
 # We could potentially enable the 2 below:
-Layout/IndentFirstHashElement:
+Layout/FirstHashElementIndentation:
```
phillipleblanc: Looks like this PR includes changes from #171 ?

jpeg-ms: Yeah it did. Not a problem anymore since #171 is merged :D

PR #173, opened 2019-12-02
Missed out an s in last PR #150

Hanstad:
The new feature using * in destination does not work because of a spelling mistake. This PR addresses this.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,10 +69,10 @@ Lint/UselessAssignment:
     - '**/spec/**/*'
 
 # We could potentially enable the 2 below:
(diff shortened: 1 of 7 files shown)


Hanstad:
Fixing issue with https://github.com/microsoft/fastlane-plugin-appcenter/pull/150

ping @jpeg-ms

jpeg-ms:
Just thinking about it now, the unit test should have caught this. Is there a chance you could look into it? The test passes but didn't verify that the destinations were correct. I don't know if it's easy to add this verification.

Hanstad:
sure, I will do yet another PR to get a unit test for this.

PR #174, opened 2019-12-03
Adding tests for retrieving all destination groups using asterik

Hanstad:
PR #150 did not test that destination groups were actually retrieved, this PR addresses that


Hanstad:
@jpeg-ms  Tests written to support PR#150

PR #185, opened 2020-01-10
Allow dsym: for all Apple builds

### LINKED ISSUE #183: dSYM is ignored when uploading Mac app
Uploading a Mac app to App Center by specifying both "file" and "dsym" options ignores the dSYM file and only uploads the app.

This seems to be due to AppcenterUploadAction only checking for an ".ipa" file extension, whereas macOS apps will be using a .zip (or possibly .pkg) extension:
https://github.com/microsoft/fastlane-plugin-appcenter/blob/8b4fa09b414ba563bea32a3a0c1db256a7dd8623/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb#L39

As a workaround one can first upload the app and then upload the dSYM separately using the "upload_dsym_only" option.

jp-andre:
Fixes #183

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -22,6 +22,13 @@ module Fastlane
     end
 
     class AppcenterUploadAction < Action
+      def self.is_apple_build(file)
+        return false unless file
+
+        file_ext = Helper::AppcenterHelper.file_extname_full(file)
+        ((Constants::SUPPORTED_EXTENSIONS[:ios] + Constants::SUPPORTED_EXTENSIONS[:mac])).include? file_ext
+      end
+
       # run whole upload process for dSYM files
       def self.run_dsym_upload(params)
         values = params.values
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -35,8 +42,9 @@ def self.run_dsym_upload(params)
...
         dsym_path = nil
         if dsym
-          # we can use dsym parameter only if build file is ipa
-          dsym_path = dsym if !file || File.extname(file) == '.ipa'
+          # we can use dsym parameter for all apple builds
+          self.optional_error("dsym parameter can only be used with Apple builds (ios, mac)") unless !file || self.is_apple_build(file)
+          dsym_path = dsym
         else
           # if dsym is not set, but build is ipa - check default path
           if file && File.exist?(file) && File.extname(file) == '.ipa'
```
Ajaykn21: can this case also be covered , where dsym is not provided in the path but we checks the package folder and considers if dsym.zip exists

spec/appcenter_upload_spec.rb
```diff
@@ -1561,6 +1561,34 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
...
+        stub_check_app(200)
+        stub_create_release_upload(200)
+        stub_upload_build(200)
+        stub_update_release_upload(200, 'committed')
+        stub_update_release(200)
+        stub_get_destination(200)
+        stub_add_to_destination(200)
+        stub_get_release(200)
+        stub_create_dsym_upload(200)
+        stub_upload_dsym(200)
```
Ajaykn21: since we are expecting the error before dsym upload, let's not stub dsym related calls.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
jp-andre: I don't understand this comment, since automatic detection is covered by the `else` case. If you still have concerns about it, let's raise a PR later :)

PR #186, opened 2020-01-12
Add action to fetch latest version and build number

### LINKED ISSUE #149: Merge AppCenter plugins
Hi guys!
Isn't be great to have other existing AppCenter plugins merged into this one? Currently, there are 3 duplicated plugins for fetching the latest release build number. 

I am also looking for a plugin, which returns `short_version` from AppCenter api response. I can create it, but it will be another AppCenter related plugin to install

jspargo:
Fixes #149

This PR adds the Fastlane actions `appcenter_fetch_version_number` to fetch the latest version number for a given app, which is returned as a string in the format `major.minor.patch.build`.

The action uses the [App Center API to list releases](https://openapi.appcenter.ms/#/distribute/releases_list), which is one of the 'distribute' endpoints. It uses the value received from the `'version'` key, though I'd be interested to see if there's an appetite to provide an action/parameter to provide the value for `'short_version'`.

### Usage
This action can be used simply by calling `appcenter_fetch_version_number`, which returns the version number as a string. For example, if your Fastfile looked like this
```
lane :test
  puts appcenter_fetch_version_number
end
```
then provided the environment variable `APPCENTER_API_TOKEN` is set, the user will be prompted to choose their app from a list (if there is more than 1) when running this lane. If it's not set, then `api_token` should be passed as a parameter.

For CI and other non-interactive use cases, the app name should be specified as a minimum, however specifying the owner name too saves on an API call:
```
lane :test
  version = appcenter_fetch_version_number(
    api_token: "my-api-token",
    owner_name: "MyAppCenterAccountName",
    app_name: "MyApplication"
  )
  puts version
end
```

### What about [`latest_appcenter_build_number`](https://github.com/jspargo/fastlane-plugin-latest_appcenter_build_number)?
To be clear, the functionality contained in this PR is identical to what currently exists in the `fastlane-plugin-latest_appcenter_build_number` plugin. If this PR is acceptable and the action within it are merged into this repository, I will put a notice in the README and a command-line-warning asking users to adopt this as the official App Center plugin and make it clear that additional features and bug fixes will only be addressed in releases from this repository.


jspargo:
Thanks @jp-andre! Just taking a look through your review, but at a glance I think it all makes sense. I might remove the valid name check altogether for the time-being - it could always be re-added across all actions in this plugin if that's something that you and the team think would be useful. What do you think?

I'm thinking perhaps the object returned could be `{id,  version, build_number }`? I think ID would be useful. One thing I don't quite follow is that in your conversion code `build_number` is being set from `version` in the API response, but `version` is the full version number including the build number (the CFBundleVersion or versionCode) - so for:
```
        "id": 1,
        "short_version": "1.0.0",
        "version": "1.0.0.1234",
```
are you thinking that the object would be `{1,  "1.0.0", "1.0.0.1234" }`?

jp-andre:
> Thanks @jp-andre! Just taking a look through your review, but at a glance I think it all makes sense. I might remove the valid name check altogether for the time-being - it could always be re-added across all actions in this plugin if that's something that you and the team think would be useful. What do you think?

Sounds good, feel free to adjust or remove as you prefer. (Note that `.` (dot) is an edge case and could be filtered out).

> I'm thinking perhaps the object returned could be `{id, version, build_number }`? I think ID would be useful. One thing I don't quite follow is that in your conversion code `build_number` is being set from `version` in the API response, but `version` is the full version number including the build number (the CFBundleVersion or versionCode) - so for:
> 
> ```
>         "id": 1,
>         "short_version": "1.0.0",
>         "version": "1.0.0.1234",
> ```
> 
> are you thinking that the object would be `{1, "1.0.0", "1.0.0.1234" }`?

I believe Android's `versionCode` would be returned as `version` and that's what we would use as `build_number`. In other words, the API should return:
` { "id": 1, "short_version": "1.0.0", "version": "1234", ... } `
And we would return ` { 1, "1.0.0", "1234" } `, though in my opinion returning a hash here would be easier to understand:
```
{ "id" => 1, "version" => "1.0.0", "build_number" => "1234" }
```

I don't know any case where we build and return a "full version" string as you mentioned. But if it's happening, I'd love to know and look into it.

Apart from these comments, could you amend the README as well to add a mention to this new action?

jspargo:
Thanks @jp-andre
> I don't know any case where we build and return a "full version" string as you mentioned. But if it's happening, I'd love to know and look into it.

I primarily work with iOS apps and have always found `version` to return the [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) and `short_version` to return [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) when using App Center upload (and back in HockeyApp days too), but perhaps this is a bug? 

Personally, I’d prefer it if the API returned what you described, but I have a feeling this is a limitation based on the way Apple handles version numbers. I’d be happy to share an example via a private channel, if that helps.

jp-andre:
> I primarily work with iOS apps and have always found `version` to return the [CFBundleVersion](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) and `short_version` to return [CFBundleShortVersionString](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) when using App Center upload (and back in HockeyApp days too), but perhaps this is a bug?

This sounds about right. Each value should correspond to exactly one entry in the build configuration files. The rest is just Apple's convention. :)

jspargo:
Hi @jp-andre - I've made the changes suggested. Changes since your last review in summary:
- Removed checking name validity (I may put a PR up for this separately in future, if we think it might be useful)
- Moved HTTP calls to helper class and follow pattern used there
- Updated README to include this new action
- Removed the interactive element of prompting for `owner_name` and `app_name` - these are now required (though Fastlane will prompt if they're missing, so that's helpful). Officially we should say that all 3 params are required.
- Changed the returned value to a hash in the format:
```
{ "id" => 1, "version" => "1.0.0", "build_number" => "1234" }
```

Looking forward to hearing back from you soon!

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
```diff
@@ -0,0 +1,99 @@
...
+        if latest_build.nil?
+          UI.user_error!("The app has no versions yet")
+          return nil
+        end
+
+        return {
+          "id" => latest_build['id'],
+          "version" => latest_build['short_version'],
+          "build_number" => latest_build['version']
+        }
```
jp-andre: This looks great. Can we document the returned struct format somewhere?

fastlane/Fastfile
```diff
@@ -193,3 +193,16 @@ lane :test do
...
+
+  UI.message("\n\n\n=====================================\n Fetch latest version: \n=====================================")
+
+  version = appcenter_fetch_version_number(
+    api_token: ENV["TEST_APPCENTER_API_TOKEN"],
+    owner_name: ENV["TEST_APPCENTER_OWNER_NAME"],
+    app_name: "MyApplication"
+  )
+
+  UI.message("#{version}")
```
jp-andre: Rubocop is complaining here. This is pretty irrelevant, to be honest, but for your information here's what I get when running `rake`:
```
fastlane/Fastfile:198:1: C: Layout/EmptyLinesAroundBlockBody: Extra empty line detected at block body beginning.
fastlane/Fastfile:207:14: C: Style/RedundantInterpolation: Prefer to_s over string interpolation.
  UI.message("#{version}")
             ^^^^^^^^^^^^

6 files inspected, 2 offenses detected
```

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -537,6 +537,12 @@ def self.get_install_url(owner_type, owner_name, app_name)
         owner_path = owner_type == "user" ? "users/#{owner_name}" : "orgs/#{owner_name}"
         return "https://install.appcenter.ms/#{owner_path}/apps/#{app_name}"
       end
+
+      # Checks that the app name is valid
+      def self.check_valid_name(name)
+        regexp = /^[a-zA-Z0-9\-]+$/i
```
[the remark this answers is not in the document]
anywherepilot: This is the regex in the model: `/^[a-zA-Z0-9-_.]+$/`

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        releases = Helper::AppcenterHelper.fetch_releases(
+          api_token: api_token,
+          owner_name: owner_name,
+          app_name: app_name
+        )
+
+        UI.abort_with_message!("No versions found for '#{app_name}' owned by #{owner_name}") unless releases
+        sorted_release = releases.sort_by { |release| release['id'] }.reverse!
+        latest_build = sorted_release.first
```
anywherepilot: Minor: Builds and releases are different things in App Center. You're dealing only with releases here.

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        if latest_build.nil?
```
anywherepilot: Defensive coding can't hurt, but is this possible? You're checking there are at least some releases above already. Or can that be an empty array?

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
```diff
@@ -0,0 +1,99 @@
...
+      def self.authors
+        ["jspargo", "ShopKeep"]
+      end
+
+      def self.run(params)
+        api_token = params[:api_token]
+        app_name = params[:app_name]
+        owner_name = params[:owner_name]
+
+        if owner_name.nil?
```
anywherepilot: Since the owner name is now required, this check and the method solving the missing value are redundant

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
jspargo: Yeh, I actually [added a spec](https://github.com/microsoft/fastlane-plugin-appcenter/pull/186/files#diff-f3feb2c2c170312dcd04d3e48f0f2611R75-R87) for this case too. Because you may have created an app but not yet uploaded any releases, you get an empty array back from the API, so I wanted to handle that as gracefully as possible. Let me know of you think this is over-engineered though?

jspargo:
Thanks @jp-andre and @anywherepilot for your reviews and helpful comments! I've addressed all the remaining comments I think.

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
anywherepilot: Since it's covering something that can happen, keep it like this.

PR #194, opened 2020-02-21
Add debug logs for all API calls in upload action

jp-andre:


jp-andre:
Ooops I don't know why there was another PR.

PR #204, opened 2020-04-16
Allow to run individual rpsecs. Fixes #203

### LINKED ISSUE #203: Can't run individual rspec test
I don't seem to be able to run individual rspec tests. It also asks for overwriting files.

```
$ rspec spec/appcenter_upload_spec.rb 
[...]
    raises an error when trying to upload a .dmg when specifying only mandatory_update (FAILED - 7)
    raises an error when trying to upload a .pkg when specifying only version (FAILED - 8)
    raises an error when trying to upload a .pkg when specifying only build_number (FAILED - 9)
    raises an error when trying to upload a .pkg when specifying only mandatory_update (FAILED - 10)
[10:51:33]: File './spec/fixtures/appfiles/app_file_empty.app.zip' already exists, do you want to override it? (y/n)
[...]
```

lacostej:


spec/appcenter_upload_spec.rb
```diff
@@ -598,7 +603,7 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
       stub_create_release_upload(200)
       stub_upload_build(200)
       stub_update_release_upload(200, 'committed')
-      stub_update_release(200)
+      stub_update_release(200, 'No changelog given')
```
jp-andre, line 606: I think it would have been much easier to change the default value in the function here :)

spec/appcenter_stub.rb
```diff
@@ -0,0 +1,20 @@
+def stub_fetch_distribution_groups(owner_name:, app_name:, groups: ["Collaborators", "test-group-1", "test group 2"])
```
jp-andre, line 1: This one is indeed shared.

PR #212, opened 2020-05-13
Add description on possible values for app_os

tedgonzalez:
Hi I just want to add this in the description so consumers can easily know the exact strings without checking the code. Let me know what you think!


jp-andre:
Thanks for this contribution!

PR #215, opened 2020-05-16
Fix appcenter_upload breaks the notarization of the macOS app #214

### LINKED ISSUE #214: appcenter_upload breaks the notarization of the macOS app
Hello,

I've found little issue with uploading of the notarized macOS app. I tried to run `appcenter_upload` with notarized _.app_ file, but when I downloaded the uploaded app, the app wasn't notarized.

I've found that the issue is with compressing of the app, which is a part of the `appcenter_upload` action. When the app is compressed without symlinks, it will break notarization of the app. So _.app_ from created _.zip_ file is no longer notarized.

eRDe33:
Hello,

I've found little issue with uploading of the notarized macOS app. I tried to run `appcenter_upload` with notarized _.app_ file, but when I downloaded the uploaded app, the app wasn't notarized.

I've found that the issue is with compressing of the app, which is a part of the `appcenter_upload` action. When the app is compressed without symlinks, it will break notarization of the app. So _.app_ from created _.zip_ file is no longer notarized.

This little adjustment of the `ZipAction` usage resolve that issue.

It fix #214 issue.

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
@@ -180,7 +180,7 @@ module Fastlane
             File.delete zip_file
           end
           UI.message("Creating zip archive: #{zip_file}")
-          file = Actions::ZipAction.run(path: file, output_path: zip_file)
+          file = Actions::ZipAction.run(path: file, output_path: zip_file, symlinks: true)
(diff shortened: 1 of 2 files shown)


lumaxis:
Thank you for the contribution @eRDe33!
It seems like the specs also need to be adjusted for this change. Would you mind looking into those, too?

Issue #115, opened 2019-09-09
Must add new apps manually to distribution group


Bjorgvin:
I have in my FastFile the following:

```
  desc "Deploy Landsbankinn Alpha to Appcenter"
  lane :appcenter do
    appcenter_upload(
      api_token: ENV["APPCENTER_TOKEN"],
      owner_type: "organization",
      owner_name: "Organization",
      app_os: "Android",
      app_platform: "Java",
      app_name: "AppName",
      apk: "./fastlane/appname.apk",
      destinations: "Internal",
      destination_type: "group"
    )
  end
```

The app does not exist in AppCenter. However, the organisation does and so does the distribution group.  When I run my command `bundle exec fastlane android appcenter` all works out fine until it is time to add the app to the given group, then I get this:

![image](https://user-images.githubusercontent.com/491852/64519100-c5dd2080-d2f3-11e9-98ed-7bf562ee16e9.png)

I then go to the AppCenter website [https://appcenter.ms/orgs/organization/people/distribution-groups/Internal/testers/apps](https://appcenter.ms/orgs/organization/people/distribution-groups/Internal/testers/apps) and add the newly created app to the distribution group manually. 

Now I run my command `bundle exec fastlane android appcenter` again and I get this:

![image](https://user-images.githubusercontent.com/491852/64519490-82cf7d00-d2f4-11e9-92b5-a27f52fd1b08.png)

So I think maybe there needs to be a default behaviour of adding new apps to distribution groups if it is not there when given a destination that is a group instead of an error saying the group is not found.

NyxReloaded:
+1 same behaviour

CaptainScavo:
+1 We would also like to see the app be automatically added if the Org & Distribution group exist

lacostej:
+1 for us as well

PR #217, opened 2020-05-18
Add newly created apps to distribution group

Hiroto-N:
Implements behaviour requested in Issue #115 

Newly created apps are not included in any distribution groups by default, so when uploading the app for the first time AppCenter by default expects the user to manually add the app to a distribution group (this is in the web UI as well). 

This PR will add the additional step of making sure that the app is added to the distribution groups provided in the 'destinations' parameter, so the app is distributed as expected.


spec/appcenter_upload_spec.rb
```diff
@@ -1373,6 +1380,8 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
       stub_get_destination(200)
       stub_add_to_destination(200)
       stub_get_release(200)
+      stub_fetch_distribution_groups(owner_name: 'owner', app_name: 'app')
```
ajaykn, line 1383: Can you please add another test case where app should have the group and skips adding it.
So that it covers this code lines
`  
  destination_names.each do |destination_name|
`
`
           unless group_names.include? destination_name
`

Hiroto-N: added!

Made sure both scenarios are checked with 2 destinations so it sends the request for one but not the other 👍

spec/appcenter_upload_spec.rb
```diff
@@ -1826,5 +1835,40 @@ def stub_add_to_destination(status, app_name = "app", owner_name = "owner", dest
...
+      stub_upload_build(200)
+      stub_update_release_upload(200, 'committed')
+      stub_update_release(200, "No changelog given")
+      stub_get_destination(200, app_name = "app", owner_name = "owner", destination_type = "group", destination_name = "Testers")
+      stub_get_destination(200, app_name = "app", owner_name = "owner", destination_type = "group", destination_name = "test-group-1")
+      stub_add_to_destination(200)
+      stub_get_release(200)
+      stub_fetch_distribution_groups(owner_name: 'owner', app_name: 'app')
+
+      should_be_callled = stub_add_new_app_to_distribution(destination_name: 'Testers')
```
ajaykn, line 1852: typo, should_be_called

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -285,6 +285,30 @@ def self.get_or_create_app(params)
         end
       end
 
+      def self.add_app_to_distribution_if_needed(params)
```
ajaykn, line 288: just to keep it consistent, please rename `add_app_to_distribution_group_if_needed`

spec/appcenter_upload_spec.rb
Hiroto-N: ah silly mistake :P - fixed

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
Hiroto-N: renamed 👍

ajaykn:
Since this is blocking the customers in automated flow of distributing, we are merging this.
In general, we are trying to avoid any new functionalities to this plugin.

ajaykn:
We have few more bug fix PR's going on. Once done, we will create a new release.

Issue #220, opened 2020-05-25
Implement upload in chunks


jp-andre:
We need to stop using the "old" HockeyApp upload flow, and follow the pattern that the portal is using. Many issues have come up over time related to timeouts during upload and recently about "file size too big".

Implementing modern upload logic, using chunks of size smaller than 256Mb.
This will potentially solve the following issues: #218 #219 #211 #137 #93.

Thanks @lumaxis for the tip.

PR #221, opened 2020-05-25
Chunk large dSYM file uploads

### LINKED ISSUE #218: Failed to upload dsym, RequestBodyTooLarge
myapp.app.dSYM.zip has 326.85 MB now.
I cannot upload dsym.

```
Error uploading dSYM 413: ﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>RequestBodyTooLarge</Code><Message>The request body is too large and exceeds the maximum permissible limit.
RequestId:db46cbf9-701e-00c0-23de-2fa0e7000000
Time:2020-05-22T02:11:06.9856835Z</Message><MaxLimit>268435456</MaxLimit></Error>
dSYM upload aborted
```

```
fastlane (2.148.1)
fastlane-plugin-appcenter (1.8.0)
```

Can I resolve this?

lumaxis:
Related to #220, fixes #218

We upload dSYMs directly to Azure Blob storage which only supports chunks with a maximum size of 256 MB, see https://docs.microsoft.com/en-us/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs#about-block-blobs
This PR adds logic that uploads files above that size limit in 100 MB chunks. The logic for this is more or less copied from some HockeyApp backend code and adjusted to work with `Faraday`.

**Note: I have not yet actually tested this with a real file!**

### DIFF
--- a/fastlane-plugin-appcenter.gemspec
+++ b/fastlane-plugin-appcenter.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'fastlane', '>= 2.96.0'
-  spec.add_development_dependency 'pry'
+  spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'rspec'
   spec.add_development_dependency 'rspec_junit_formatter'
--- a/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
+++ b/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
@@ -1,7 +1,14 @@
+class File
+  def each_chunk(chunk_size = 1_000_000)
+    yield read(chunk_size) until eof?
+  end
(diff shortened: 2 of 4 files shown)


jp-andre:
I've tested this with fake files and can't upload mapping files for some reason. This is what I get:
```
Error uploading mapping 400: ﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidXmlDocument</Code><Message>XML specified is not syntactically valid.
RequestId:a252ce16-101e-0095-3635-33c249000000
Time:2020-05-26T08:14:47.5632091Z</Message><LineNumber>0</LineNumber><LinePosition>0</LinePosition><Reason /></Error>
```
The content is exactly the same as in the browser (using 4Mb chunks), so there's something strange happening in server side, or in the ruby encoding side. I tried but failed to sniff the outgoing traffic when running the command. It would be great to see if there's indeed some invalid characters in the body.

jp-andre:
Closing this now. Would love to see this actually work, so please re-open if you have a fix.

PR #222, opened 2020-05-26
Adds support for EDDSA signatures

mz2:
Adds support for EDDSA signing Sparkle builds.

### DIFF
--- a/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
+++ b/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
@@ -409,12 +409,13 @@ module Fastlane
       end
 
       # updates release metadata
-      def self.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature)
+      def self.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature, eddsa_signature)
         return if dsa_signature.to_s == ''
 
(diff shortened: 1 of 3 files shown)


jp-andre:
Thanks for your contribution @mz2. Please make sure that the spec tests pass.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -409,12 +409,13 @@ def self.update_release(api_token, owner_name, app_name, release_id, release_not
       end
 
       # updates release metadata
-      def self.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature)
+      def self.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature, eddsa_signature)
         return if dsa_signature.to_s == ''
```
ajaykn, line 413: I think, this condition needs to be updated with eddsa_signature as well.

jp-andre:
@mz2 if you are still interested in getting this PR merged, please consider the above comments.

kkyryl:
@jp-andre, @ajaykn I have submitted a similar pull request which accounts for your comments above. Please review if possible since it's blocking us from using App Center (we are migrating from a self hosted solution and we need EdDSA to work)

DmitriyKirakosyan:
Closing as stale.

PR #226, opened 2020-06-09
Add new option to appcenter_fetch_version_number to fetch a specific version

Qutaibah:
Getting the last build number for a specific version, this functionality is already available for TestFlight, and for me it's one of the most important features to automate version and build numbers using Fastlane

I'm not a ruby developer so no tests 😅 if someone can help with this, it would be perfect, or I'll do it after I learn it


jp-andre:
Hey @Qutaibah, thanks for your contribution.

How is this different from the existing action `appcenter_fetch_version_number`?
If I understand correctly, this is filtering on a specific `version` to get the `build_number` from it. Why not add an optional parameter to the existing action `appcenter_fetch_version_number`?

Please avoid Apple-specific naming such as train, as this plugin also supports Google Play and groups distribution with the same parameter names.

Once we agree on the base code, please include the tests. Usually they're fairly easy to write by using appropriate copy & paste of the existing ones ;)

Qutaibah:
@jp-andre Thanks for your reply, I updated the following
1- added the functionality as an option to `appcenter_fetch_version_number` (don't know why i didn't think of that 🤔 )
2- replaced the train word.
3- for tests, sure I'll do that once we're ok

Thanks

lib/fastlane/plugin/appcenter/actions/appcenter_fetch_version_number.rb
```diff
@@ -1,45 +1,58 @@
...
+
+        latest_release = sorted_releases.last
 
         if latest_release.nil?
-          UI.user_error!("This app has no releases yet")
+          if version.nil?
+            UI.user_error!("This app has no releases yet")
+            return nil
+          end
+          UI.user_error!("The provided version has no releases yet")
```
jp-andre, line 47: Maybe also print out the requested version.

Qutaibah: Done

Qutaibah:
Hi @jp-andre, just in case you missed it, I did what we agreed on

PR #230, opened 2020-06-26
Use the new appcenter endpoints for release upload

guperrot:
As upload proxy is deprecated.


guperrot:
Missing:
* error cases on 200 for upload API
* error status patching not working
* splitting in chunks is mandated by server

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -186,14 +203,28 @@ def self.run_release_upload(params)
...
+          UI.abort_with_message!("Upload aborted") unless uploaded
+
+          UI.message("Finishing release...")
+          finish_url = "#{upload_details['upload_domain']}/upload/finished/#{upload_details['package_asset_id']}?token=#{upload_details['url_encoded_token']}"
+          finished = Helper::AppcenterHelper.finish(finish_url, api_token, owner_name, app_name, upload_id, timeout)
+          UI.abort_with_message!("Upload aborted") unless finished
 
-          if uploaded
-            release_id = uploaded['release_id']
+          release_status_url = "v0.1/apps/#{owner_name}/#{app_name}/uploads/releases/#{upload_id}"
```
jp-andre, line 224: `UI.message("Waiting for release to be ready...")` or something like that

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -229,50 +234,119 @@ def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, sym
...
+      end
+
+      # Verifies a successful upload to App Center
+      # returns:
+      # successful upload response body.
+      def self.finish(finish_url, api_token, owner_name, app_name, upload_id, timeout)
+        connection = self.connection(finish_url)
 
+        UI.message("DEBUG: POST #{finish_url}") if ENV['DEBUG']
+        UI.message("DEBUG: POST body <data>\n") if ENV['DEBUG']
```
jp-andre, line 277: I see no data being sent through this POST request. Is this really correct?

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -229,50 +234,119 @@ def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, sym
...
         end
       end
 
+      # upload binary for specified upload_url
+      # if succeed, then commits the release
+      # otherwise aborts
+      def self.upload_build(api_token, owner_name, app_name, file, upload_id, upload_url, content_type, chunk_size, timeout)
+        block_number = 1
+
+        File.open(file).each_chunk(chunk_size) do |chunk|
```
jp-andre, line 308: This logic seems pretty "basic" here. For the CLI, we have auto-retry and parallel uploads.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -229,50 +234,119 @@ def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, sym
...
+          upload_chunk_url = "#{upload_url}&block_number=#{block_number}"
+          connection = self.connection(upload_chunk_url, true)
+
+          UI.message("DEBUG: POST #{upload_chunk_url}") if ENV['DEBUG']
+          UI.message("DEBUG: POST body <data>\n") if ENV['DEBUG']
+          response = connection.post do |req|
+            req.options.timeout = timeout
```
jp-andre, line 315: This changes the definition of timeout, since we now have `timeout * chunk_count` in total.

I see two use cases for timeout, before this PR:
- Increase it so that large uploads don't fail.
- Reduce it to avoid getting stuck in a user's CI pipeline, when App Center is slow to accept uploads for whatever reason.

I think keeping `timeout` the way you've done it here is fine, but we should be clear that it's different from what we used to have. We already upload symbols (and `timeout` does not work for them, unfortunately) as well as builds, and anecdotal evidence points to the first use case above. But I think this warrants a note in the documentation & README.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -186,14 +203,28 @@ def self.run_release_upload(params)
...
         if upload_details
-          upload_id = upload_details['upload_id']
-          upload_url = upload_details['upload_url']
+          upload_id = upload_details['id']
+          
+          UI.message("Setting Metadata...")
+          content_type = Constants::CONTENT_TYPES[File.extname(file)&.delete('.').to_sym] || "application/octet-stream"
+          set_metadata_url = "#{upload_details['upload_domain']}/upload/set_metadata/#{upload_details['package_asset_id']}?file_name=#{File.basename(file)}&file_size=#{File.size(file)}&token=#{upload_details['url_encoded_token']}&content_type=#{content_type}"
+          chunk_size = Helper::AppcenterHelper.set_metadata(set_metadata_url, api_token, owner_name, app_name, upload_id, timeout)
+          UI.abort_with_message!("Upload aborted") unless chunk_size
```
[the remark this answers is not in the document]
owlstronaut: Just the file size. There is a maximum file size per chunk, so the chunks are maxed out until the upload is finished. The last chunk is the only one that should be smaller than the max. I think the max is 4mb

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
owlstronaut: nope, removing.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -229,50 +234,119 @@ def self.upload_symbol(api_token, owner_name, app_name, symbol, symbol_type, sym
...
       # Commits or aborts the upload process for a release
       def self.update_release_upload(api_token, owner_name, app_name, upload_id, status)
         connection = self.connection
 
-        url = "v0.1/apps/#{owner_name}/#{app_name}/release_uploads/#{upload_id}"
+        url = "v0.1/apps/#{owner_name}/#{app_name}/uploads/releases/#{upload_id}"
         body = {
-          status: status
+          upload_status: status,
+          id: upload_id
```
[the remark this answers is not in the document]
owlstronaut: Yeah it is 🤷

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
guperrot: There are also many steps involved in creating a release, which translates into several API calls before uploading the chunk itself, so I don't think it was correct even before.

We could make the timeout more global and remember time left for the entire process.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
guperrot: I don't think parallel is in scope for this feature. We will discuss retries.

guperrot: We added a few retries here

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
guperrot: Added a note on request timeout in the readme.

PR #231, opened 2020-07-09
Revert ruby version file

guperrot:
It was accidentally changed while testing locally #230 with rbenv instead of using rvm


jp-andre:
You know what? Let's keep it at 2.6.5. I also switched to `rbenv` myself 😅 Bumping it for next 2.6.x or next 2.7.x (once fastlane is fully ready for it) is a tiny change.

guperrot:
Alrighty then

Issue #246, opened 2020-09-29
Cant upload android apk via appcenter_upload with 1.9.0


filipkowicz:
from what i see backend respond with 200 and empty body for [source](https://github.com/microsoft/fastlane-plugin-appcenter/blob/bf0b56f5d4d0d24017f600976b8c3335dd2da84f/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb#L251) call

action is failing with "Error uploading binary" and "Upload aborted"
for 1.8.1 with same params works well

filipkowicz:
side note - add 
` ENV['DEBUG'] = "true"`
before calling action

joseignacioojeda:
I am having the same problem

filipkowicz:
I've give up and switch to 1.8.1

lampietti:
same here, and it was working fine yesterday before the last incident https://status.appcenter.ms/ ...

leo-pedidosya:
Same here

PatrykKaczmarek:
Same here. Interestingly building (and uploading binary) the same project with exactly the same configuration passes when using bitrise CLI locally instead of building it on their servers.

sedwo:
Been struggling with this all evening.  I concur, `fastlane-plugin-appcenter (1.9.0)` breaks the upload process.

![image](https://user-images.githubusercontent.com/1434449/94628033-03a94000-028d-11eb-8346-f651ac0c364c.png)

NirGrocerKey:
having the same issues since the last incident, any help how to roll back to 1.8.1?

b3hni4:
I have the same issue, workaround is to use fastlane 2.160.0 until this is fixed

alexv-anyvision:
Having same issue

neil-gebbie-smarterley:
This is a duplicate of https://github.com/microsoft/fastlane-plugin-appcenter/issues/247 - for the time being you can lock your version to 1.8.1 by following these instructions - https://docs.fastlane.tools/plugins/using-plugins/#plugin-source

`gem 'fastlane-plugin-appcenter', "1.8.1"`

ff-ahumberston:
Having attempted the version local to 1.8.1, we are now seeing the following error:

`INFO [2020-09-30 11:34:12.10]: ------------------------------
INFO [2020-09-30 11:34:12.10]: --- Step: appcenter_upload ---
INFO [2020-09-30 11:34:12.10]: ------------------------------

[...]

PR #248, opened 2020-10-02
Specify content-type on chunk upload api since the default is not supported

### LINKED ISSUE #247: Cannot upload ios via appcenter_upload using 1.9.0
I am using the following command:
```
    appcenter_upload(
        api_token: ENV["APPCENTER_API_TOKEN"],
        owner_name: ENV["APPCENTER_OWNER_NAME"],
        app_name: ENV["APPCENTER_APP_NAME"],
        file: "PATH-To-IPA",
        release_notes: "RELEASE-NOTES"
        )
````
and it is giving the following error:
```
Client error: 400: {"error"=>{"code"=>400, "message"=>"The request contains invalid content."}}
```

It is working fine with the 1.8.1 version though.

### PR DESCRIPTION
**Pull request by rimarsh**
After updating fastlane to 2.161.0 or later, there is a failure when uploading builds using this plugin. The issue appears to be caused by [this change](https://github.com/fastlane/fastlane/commit/7a6b9d684b09c71279daf3471a44bce53a020a53) which altered default behavior for `Content-Type` on requests by any gem loaded by fastlane. As a result, this upload request is now issued with `application/www-x-form-urlencoded` instead of having no `Content-Type` header. Specifying a desired header fixes this issue, though I'm not sure if `application/octet-stream` is correct for this request.

The error seen currently is:
```
Client error: 400: {"error"=>{"code"=>400, "message"=>"The request contains invalid content."}}
```

Fixes #247 and #246

### DIFF
--- a/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
+++ b/lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
@@ -324,6 +324,7 @@ module Fastlane
                 req.options.timeout = timeout
                 req.headers['internal-request-source'] = "fastlane"
                 req.headers['Content-Length'] = chunk.length.to_s
+                req.headers['Content-Type'] = 'application/octet-stream'
                 req.body = chunk
               end
               UI.message("DEBUG: #{response.status} #{JSON.pretty_generate(response.body)}\n") if ENV['DEBUG']


vanlooverenkoen (approved):

DDamian (approved):

lampietti (approved):

uwemeier-sclable:
Why is this not merged already?

vanlooverenkoen:
@jp-andre is it possible to take a look at this? This is a blocking pull request.

NanoTribe:
This would help us a great deal, please merge if it works well.

joshdholtz:
This is the correct solution! We removed a hack for default setting the content type in all requests (because bad) within fastlane in 2.161.0 and it looks like some other plugins were depending on that 🤷‍♂️ 

I approve this! Can we get this merged?

joshdholtz (approved):
This is a good fix! This was a side effect of https://github.com/fastlane/fastlane/pull/17218

LGTM 🔥

lumaxis (approved):
Would be nice to have specs for this but let's merge it to get people unblocked 🙏🏼

rimarsh:
Thanks @lumaxis.

Happy to help with the specs. I wasn't sure the best way to go about writing them since it would just be mocking the current observed behavior of that endpoint. Is there an official doc for that API which specifies the supported content types?

PR #250, opened 2020-10-07
Enable support fo EdDSA signatures in addition to DSA

kkyryl:
Both are supported by the App Center but only DSA is currently supported by the fastlane plugin


lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -546,15 +546,23 @@ def self.update_release(api_token, owner_name, app_name, release_id, release_not
...
+      def self.update_release_metadata(api_token, owner_name, app_name, release_id, dsa_signature = '', ed_signature = '')
+        return if dsa_signature.to_s == '' && ed_signature.to_s == ''
 
         url = "v0.1/apps/#{owner_name}/#{app_name}/releases/#{release_id}"
-        body = {
-          metadata: {
-            dsa_signature: dsa_signature
-          }
-        }
+        if dsa_signature.to_s != ''
```
kkyryl, line 553: In my experience, those are not used together therefore this check is either/or.

jp-andre: Agreed. You can enforce that with `conflicting_options` and `conflict_block` in the Options list. Or alternatively allow passing both like #222 was doing.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
```diff
@@ -143,6 +143,7 @@ def self.run_release_upload(params)
         build_number = params[:build_number]
         version = params[:version]
         dsa_signature = params[:dsa_signature]
+        ed_signature = params[:ed_signature]
```
jp-andre, line 146: `eddsa` would be easier to understand in my opinion.

jp-andre (commented):
Looks good but this is lacking tests (specs). Would love to merge if we get new tests.
Also we need to update README to add the new field.

lib/fastlane/plugin/appcenter/actions/appcenter_upload_action.rb
kkyryl: I agree - but I was following the naming convention in the API and in the API the field is ed_signature. I can go either way, wasn't sure what's community preference here.

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
kkyryl: @jp-andre Updated to allow both values.

kkyryl:
@jp-andre Updated per your comments:
- Added support for both values
- Updated README
- Updated spec

kkyryl:
@jp-andre any chance we can move this forward?

lib/fastlane/plugin/appcenter/helper/appcenter_helper.rb
```diff
@@ -546,15 +546,20 @@ def self.update_release(api_token, owner_name, app_name, release_id, release_not
...
-          }
+          metadata: {}
         }
+        
+        if dsa_signature.to_s != ''
+          body["metadata"]["dsa_signature"] = dsa_signature
+        end
+        if ed_signature.to_s != ''
+          body["metadata"]["ed_signature"] = ed_signature
+        end
```
jp-andre, line 562: This code is not valid. `metadata` in `body` is a symbol (`:metadata`), you're indexing here as a string `"metadata"`.
```suggestion
        if dsa_signature.to_s != ''
          body[:metadata]["dsa_signature"] = dsa_signature
        end
        if ed_signature.to_s != ''
          body[:metadata]["ed_signature"] = ed_signature
        end
```

jp-andre (changes requested):
Sorry, the spec tests don't pass. I've added a change (how was I allowed to write in your fork? I wonder). Please kindly fix the tests and ping me again if I forget about this PR.

kkyryl:
@jp-andre thank you for helping with the code, appreciate it. I fixed the issue with the spec and adjusted, looks like it passes now.

kkyryl:
@jp-andre just pinging to see if we can get this going

jp-andre (approved):

anywherepilot (approved):

PR #319, opened 2023-06-18
[Add] support hermes in code push

phu-tang-tiki:
I see that the appcenter codepush cli currently support hermes on codepush. So I make this PR to add this parameter on fastlane function. 

Please review and let me know if need to improve

### DIFF
--- a/lib/fastlane/plugin/appcenter/actions/appcenter_codepush_release_react.rb
+++ b/lib/fastlane/plugin/appcenter/actions/appcenter_codepush_release_react.rb
@@ -75,8 +75,8 @@ module Fastlane
         if xcode_project_file
           command += "--xcode-project-file #{xcode_project_file} "
         end
-        if use_hermes
-          command += "--use-hermes"
+        unless use_hermes.nil?
+          command += "--use-hermes #{use_hermes} "
         end
         if dry_run
           UI.message("Dry run!".red + " Would have run: " + command + "\n")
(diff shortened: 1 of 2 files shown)


DmitriyKirakosyan:
@phu-tang-tiki , thank you for the contribution!
The argument `use-hermes` is used to bypass the hermes check. In normal scenario you don't need to use this argument, and hermes will be automatically enabled if it is set in your project configuration. However in some cases you would need to bypass the check due to specific project configuration. Is it the case why you are adding this parameter?

phu-tang-tiki:
yes, I need to use that parameter, specially for Android version. Let say we custom a lot in build flavor and variants and have to enable it manually.

lib/fastlane/plugin/appcenter/actions/appcenter_codepush_release_react.rb
```diff
@@ -74,6 +75,9 @@ def self.run(params)
         if xcode_project_file
           command += "--xcode-project-file #{xcode_project_file} "
         end
+        if use_hermes
+          command += "--use-hermes"
```
DmitriyKirakosyan, line 79:
```suggestion
          command += "--use-hermes "
```

DmitriyKirakosyan: What if we wan't to bypass disabling hermes? I think it should be
```
command += "--use-hermes #{use_hermes} "
```

phu-tang-tiki: you're right, it should be `--use-hermes #{use_hermes}` let me update

phu-tang-tiki: added check and add some test for sure

phu-tang-tiki:
@microsoft-github-policy-service agree

DergachevE (approved):

dungpasoftware:
> yes, I need to use that parameter, specially for Android version. Let say we custom a lot in build flavor and variants so I have to enable it manually.

do we have parameters to custom flavor for android? I try to use appcenter_codepush_release_react, but don't find any parameter to custom flavor for code push

PR #328, opened 2023-10-19
fix(local cli): prevent npm from parsing arguments

LouisKraemer:
Prevent npm from parsing arguments. Without this, devs cannot use local cli.

https://docs.npmjs.com/cli/v7/commands/npm-exec#npx-vs-npm-exec


LouisKraemer:
@AnatolyPristensky Do you know why the CI is not reporting back?

Priskorn (approved):

